PR #72, opened 2021-05-14
MC-9476 Term label improvement

adjl:


mdm-plugin-terminology/grails-app/utils/uk/ac/ox/softeng/maurodatamapper/terminology/bootstrap/BootstrapModels.groovy
```diff
@@ -104,94 +104,99 @@ class BootstrapModels {
...
+        terminology.addToTerms(top)
+        List<Term> terms = [top]
+        (1..101).each {
+            Term t = new Term(createdBy: DEVELOPMENT, code: "CTT$it", definition: "Complex Test Term $it")
+            terminology.addToTerms(t)
+            terms += t
+        }
+
+        terms[101].tap {
+            code = definition
```
olliefreeman, line 140: slight niggle, as we have all the codes for this example being nice and neat can we swop this so the definition = code...then when we're using this for UI and testing we get CTT### otherwise it risks falling over on other things we dont expect.

olliefreeman (changes requested):
Pretty much all good apart from the bootstrapping sample...please change this and it will then require a few of the functional tests to be changed

olliefreeman (approved):
Looks good

PR #78, opened 2021-05-19
MC-9481 Add canAddRule to list of availableActions for Editors

adjl:


olliefreeman:
In hindsight this is actually an uneeded action as rules are facets which can only be added when the `update` action is available, there is no need for this action

PR #85, opened 2021-06-14
Feature/mc 9526

jamesrwelch:
Update to work on Folders; various bug fixes; add more tests

### DIFF
--- a/mdm-plugin-profile/grails-app/services/uk/ac/ox/softeng/maurodatamapper/profile/ProfileService.groovy
+++ b/mdm-plugin-profile/grails-app/services/uk/ac/ox/softeng/maurodatamapper/profile/ProfileService.groovy
@@ -108,20 +108,15 @@ class ProfileService {
         profileProviderService.storeProfileInEntity(multiFacetAwareItem, profile, user)
     }
 
-    def validateProfile(ProfileProviderService profileProviderService, MultiFacetAware multiFacetAwareItem, HttpServletRequest request, User user) {
+    def validateProfile(ProfileProviderService profileProviderService, HttpServletRequest request) {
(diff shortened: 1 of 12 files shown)


mdm-plugin-profile/grails-app/services/uk/ac/ox/softeng/maurodatamapper/profile/ProfileService.groovy
```diff
@@ -105,10 +105,10 @@ class ProfileService {
 
           */
 
-        profileProviderService.storeProfileInEntity(catalogueItem, profile, user)
+        profileProviderService.storeProfileInEntity(multiFacetAwareItem, profile, user)
     }
 
-    def validateProfile(ProfileProviderService profileProviderService, CatalogueItem catalogueItem, HttpServletRequest request, User user) {
+    def validateProfile(ProfileProviderService profileProviderService, MultiFacetAware multiFacetAwareItem, HttpServletRequest request, User user) {
```
olliefreeman, line 111: i dont see the reason for this parameter..its unused
Also it would be much better and cleaner to use Grails to bind the request directly into a List of ProfileSections rather than custom buidling it?

olliefreeman (changes requested):
Just the unused parameter in the validate service which makes no sense

mdm-plugin-profile/grails-app/services/uk/ac/ox/softeng/maurodatamapper/profile/ProfileService.groovy
jamesrwelch: I've removed the parameter for now - I think it'll be re-instated when I complete the next iteration of profiles with derived values.  I've fixed the method so it now works correctly.

PR #87, opened 2021-06-15
Feature/mc 9537

OButlerOcc:
Made use of the command object to store the new label as we can later expand upon it
The object uses the valadatable interface to enforce not null and not empty checking because, again we can extend it.
The label rename takes place in the catalogueitemservice so it can be used for other catalogue item.
The json for editing a lable is sending [copyLabel :  ] with the post


olliefreeman (approved):
Looks good, once the tests pass i'll merge this in

olliefreeman:
test failure was a redherring for a date which is not parsable inside the test env

PR #107, opened 2021-07-15
Authority tests complete

OButlerOcc:
Seems even the admin user can't get
a show on the /UUID endpoint.
I've verified a find works on the same UUID. Commented out for now till fixed 
or I figure out something wrong with the test


olliefreeman (changes requested):
I have updated the security in the interceptor to fix the permissions please update the tests accordingly

* authenticated users can read and list
* no-one can edit

olliefreeman:
because of the security changes i made, can you look at `AuthorityInterceptorSpec` and fix the broken tests in there, change them to what they are coming out as now, do not change the actual interceptor

olliefreeman:
closing this as i merged the work into the mc-9499 branch

PR #120, opened 2021-07-28
Added new actions finalisedReadAction and finalisedEditAction

OButlerOcc:
Fixed tests that broke due to new actions


olliefreeman (changes requested):
Update action names

olliefreeman (approved):

PR #126, opened 2021-08-03
MC-9497 ReferenceDataValue functional specs

adjl:


adjl:
There is a strange bug causing the failures in mdm-plugin-referencedata/ReferenceDataModelFunctionalSpec... All of its tests pass when you run the spec on its own. But when running all tests in the plugin altogether (`gradle :mdm-plugin-referencedata:integrationTest`) it sometimes fails by returning the `referenceDataModel.authority.url` as `https://localhost` instead of `http://localhost`.

Any idea as to the cause? (Flaky test perhaps?)

Regardless, I can't see at a glance how the bug is caused by this PR.

olliefreeman:
@a

> There is a strange bug causing failures in mdm-plugin-referencedata/ReferenceDataModelFunctionalSpec... All of its tests pass when you run the spec on its own. But when running all tests in the plugin altogether (`gradle :mdm-plugin-referencedata:integrationTest`), it sometimes fails by returning the `referenceDataModel.authority.url` as `https://localhost` instead of `http://localhost`.
> 
> Any idea as to the cause? (Flaky test perhaps?)
> 
> Regardless, I can't see at a glance how the bug is caused by this PR.

You can't run all the integrationTests at the same time as that runs the IT and FT tests.
You either need to run them like the `./run-all-tests.sh` does using gradle
or do as the README instructs and use grails CLI

then 
* `test-app -Dgradle.integrationTest=true -integration` runs the IT 
* `test-app -Dgradle.functionalTest=true -integration` runs the FT

this will probably solve the issue you're finding.

mdm-plugin-referencedata/src/integration-test/groovy/uk/ac/ox/softeng/maurodatamapper/referencedata/provider/XmlReferenceDataImporterExporterServiceSpec.groovy
```diff
@@ -410,10 +404,8 @@ class XmlReferenceDataImporterExporterServiceSpec extends BaseReferenceDataModel
 
     void 'RDM09: test reference data import and export'() {
         given:
-        setupData()
-
-        expect:
-        ReferenceDataModel.count() == 2
+        // Do not load test Reference Data Models because the import is a duplicate and breaks Reference Data Value uniqueness
```
olliefreeman, line 407: if the import breaks the uniqueness then there is another issue, as we should be able to import the same model multiple times with different branch names, therefore we got the unique constraint wrong

mdm-plugin-referencedata/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/referencedata/item/ReferenceDataValueController.groovy
```diff
@@ -32,6 +33,42 @@ class ReferenceDataValueController extends EditLoggingController<ReferenceDataVa
         super(ReferenceDataValue)
     }
 
+    @Override
```
olliefreeman, line 36: protected methods should be below public ones please

mdm-plugin-referencedata/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/referencedata/item/ReferenceDataValue.groovy
```diff
@@ -40,7 +40,7 @@ class ReferenceDataValue implements CreatorAware, Diffable<ReferenceDataValue> {
     static belongsTo = [referenceDataModel: ReferenceDataModel, referenceDataElement: ReferenceDataElement]
 
     static constraints = {
-        value blank: true, nullable: true
+        value blank: true, nullable: true, unique: 'rowNumber'
```
olliefreeman, line 43: we obv got this wrong, you probably want to add `unique: 'rowNumber,referenceDataElement'`

mdm-plugin-referencedata/grails-app/services/uk/ac/ox/softeng/maurodatamapper/referencedata/item/ReferenceDataValueService.groovy
```diff
@@ -62,7 +62,6 @@ class ReferenceDataValueService implements DomainService<ReferenceDataValue> {
 
     void delete(ReferenceDataValue referenceDataValue, boolean flush = false) {
         if (!referenceDataValue) return
```
olliefreeman, line 64: why have you removed this??

olliefreeman (changes requested):
We need to review the uniqueness constraint.

We may need to implement a complicated validator but we may be able to get away with the change i propose

mdm-plugin-referencedata/grails-app/services/uk/ac/ox/softeng/maurodatamapper/referencedata/item/ReferenceDataValueService.groovy
adjl: `breadcrumbTree` is not (no longer?) a property of `ReferenceDataValue`, causing an exception. If there is an alternate way of achieving this, I could comment the statement out and return to it in a separate PR.

mdm-plugin-referencedata/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/referencedata/item/ReferenceDataValueController.groovy
adjl: Done

mdm-plugin-referencedata/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/referencedata/item/ReferenceDataValue.groovy
adjl: This change causes `BeanCreationException`s and `No GORM implementations configured. Ensure GORM has been initialized correctly`. :/

For the sake of exhaustion, `uniqueness: rowNumber, referenceDataModel` and `uniqueness: rowNumber, referenceDataElement, referenceDataModel` are no good too.

mdm-plugin-referencedata/src/integration-test/groovy/uk/ac/ox/softeng/maurodatamapper/referencedata/provider/XmlReferenceDataImporterExporterServiceSpec.groovy
adjl: For reference, importing with `importAsNewBranchModelVersion = true` still fails and breaks uniqueness.

mdm-plugin-referencedata/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/referencedata/item/ReferenceDataValue.groovy
adjl: Silly error. Constraining on two or more fields requires passing them as a list. 🤦‍♀️ `unique: ['rowNumber', 'referenceDataElement']` works.

mdm-plugin-referencedata/src/integration-test/groovy/uk/ac/ox/softeng/maurodatamapper/referencedata/provider/XmlReferenceDataImporterExporterServiceSpec.groovy
adjl: Fixed with the updated `unique` constraint.

olliefreeman (approved):
looks good

PR #128, opened 2021-08-03
Don't supply empty username / password params to email service

### LINKED ISSUE #127: Allow anonymous email server connections
An anonymous SMTP server doesn't need username / password submitted to it - in fact it may refuse the connection if any credentials are supplied - even empty strings.
If email username / password are not supplied, this should be treated as an anonymous server and so the parameters should not be part of the Mailer connection configuration.

jamesrwelch:
And add extra logging in the event of failure.
Hopefully resolves #127

### DIFF
--- a/mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/provider/email/BasicEmailProviderService.groovy
+++ b/mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/provider/email/BasicEmailProviderService.groovy
@@ -37,7 +37,7 @@ class BasicEmailProviderService extends EmailProviderService {
         allProps = properties
         simpleMailProps = new Properties()
         for (ConfigLoader.Property p : ConfigLoader.Property.values()) {
-            if (properties[p.key()]) {
+            if (properties[p.key()] && properties[p.key()] != "" ) {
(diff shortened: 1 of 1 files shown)


mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/provider/email/BasicEmailProviderService.groovy
```diff
@@ -37,7 +37,7 @@ class BasicEmailProviderService extends EmailProviderService {
         allProps = properties
         simpleMailProps = new Properties()
         for (ConfigLoader.Property p : ConfigLoader.Property.values()) {
-            if (properties[p.key()]) {
+            if (properties[p.key()] && properties[p.key()] != "" ) {
```
olliefreeman, line 40: this is unnecessary as groovy string null == ''
if the system is trying to authenticate then its not because of this code.
We need to do some proper testing to work out whats going on

olliefreeman (changes requested):
We need to test why its authenticating, or is the Imperial system using details it should have set to null.
These suggested changes wont have any effect

olliefreeman:
Im closing the PR as this solution will not solve the issue as its doing whats already being done, we need to find a way to test the issue before attempting a solution

PR #136, opened 2021-08-24
Added TreeItem Endpoint addition to show under the /ancestors option

OButlerOcc:
The endpoint works its way up from the provided catalogue item till it reaches the top of the tree then wraps
all the proceding objects in the top level container.
changes were made to the tree item controller, service, gson view and functional spec.
Some JDOC explanation has been provided for the recursive implementation because it could be unclear


olliefreeman:
closing as needs rebasing onto develop and also branch name is wrong

PR #141, opened 2021-08-27
Feature/gh 131

gammonpeter:


mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/provider/exporter/ExportModel.groovy
```diff
@@ -29,6 +30,7 @@ class ExportModel {
     String modelExportTemplatePath
     String exportModelType
     Namespace modelXmlNamespace
+    Path path
```
olliefreeman, line 33: This shouldnt be here, the ExportModel object has no path. Please remove

mdm-common/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/path/Path.groovy
```diff
@@ -248,4 +248,16 @@ class Path {
     static boolean isValidPath(String possiblePath) {
         from(possiblePath).toString() == possiblePath
     }
+
+    static Path toPathPrefix(CreatorAware domain, String prefix)
+    {
+        List<CreatorAware> paths = [];
+        paths.push(domain)
+
+        while (((CreatorAware) paths.first()).getPathPrefix() != prefix) {
```
olliefreeman, line 257: Unnecessary cast here as all objects in the path are creator aware.

mdm-common/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/path/Path.groovy
```diff
@@ -248,4 +248,16 @@ class Path {
...
+             paths.push(paths.first().getPathParent())
```
olliefreeman, line 258: This needs a check, probably up in the while statement `paths.first().getPathPrefix() != prefix && Utils.parentClassIsAssignableFromChild(PathAware, paths.first())`

you need to confirm the paths.first() actually has the method `getPathParent()`

This will require you to move PathAware into mdm-common which is fine

mdm-common/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/path/Path.groovy
```diff
@@ -248,4 +248,16 @@ class Path {
     static boolean isValidPath(String possiblePath) {
         from(possiblePath).toString() == possiblePath
     }
+
+    static Path toPathPrefix(CreatorAware domain, String prefix)
+    {
+        List<CreatorAware> paths = [];
```
olliefreeman, line 254: Can we rename this to `objectsInPath`  or something similar ... to me a variable called paths is of type `Path`

olliefreeman (changes requested):
All okay apart from the stuff in the Path class, can you make the requested changes around the new method and the casting please

mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/provider/exporter/ExportModel.groovy
gammonpeter: Updated

olliefreeman (approved):

PR #142, opened 2021-08-31
Added TreeItem Endpoint addition to show under the /ancestors option

olliefreeman:
The endpoint works its way up from the provided catalogue item till it reaches the top of the tree then wraps
all the proceding objects in the top level container.
changes were made to the tree item controller, service, gson view and functional spec.
Some JDOC explanation has been provided for the recursive implementation because it could be unclear

### DIFF
--- a/mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/UrlMappings.groovy
+++ b/mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/UrlMappings.groovy
@@ -247,6 +247,7 @@ class UrlMappings {
                     get '/'(controller: 'treeItem', action: 'index')
                     get "/${containerId}"(controller: 'treeItem', action: 'show')
                     get "/${catalogueItemDomainType}/$catalogueItemId"(controller: 'treeItem', action: 'show')
+                    get "/${catalogueItemDomainType}/$catalogueItemId/ancestors"(controller: 'treeItem', action: 'ancestors')
                     get "/search/$searchTerm"(controller: 'treeItem', action: 'search')
                 }
                 "/full/$modelDomainType/$modelId"(controller: 'treeItem', action: 'fullModelTree')
--- a/mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/tree/TreeItemController.groovy
+++ b/mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/tree/TreeItemController.groovy
@@ -132,6 +132,16 @@ class TreeItemController extends RestfulController<TreeItem> implements MdmContr
         respond(treeItemService.buildFullModelTree(model))
     }
 
+    def ancestors() {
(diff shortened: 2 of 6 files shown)


mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/UrlMappings.groovy
```diff
@@ -249,6 +249,7 @@ class UrlMappings {
                     get '/'(controller: 'treeItem', action: 'index')
                     get "/${containerId}"(controller: 'treeItem', action: 'show')
                     get "/${catalogueItemDomainType}/$catalogueItemId"(controller: 'treeItem', action: 'show')
+                    get "/${catalogueItemDomainType}/$catalogueItemId/$ancestors"(controller: 'treeItem', action: 'show')
```
olliefreeman, line 252: This is not necessary, the param should be `includeAncestors` and it would be passed in as a query param `?includeAncestors=true`

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/tree/TreeItemController.groovy
```diff
@@ -73,6 +73,10 @@ class TreeItemController extends RestfulController<TreeItem> implements MdmContr
         CatalogueItem catalogueItem = treeItemService.findTreeCapableCatalogueItem(params.catalogueItemClass, params.catalogueItemId)
         if (!catalogueItem) return notFound(CatalogueItem, params.catalogueItemId)
 
+        if (params.ancestors) {
```
olliefreeman, line 76: param should be `includeAncestors` and you'll need `params.boolean('includeAncestors', false)` to check boolean

olliefreeman (commented):
You have forgotten to actually add the content of the CI which was requested.

The idea was to return the usual content but wrapped up with the ancestors. 

I would suggest let the service method `buildCatalogueItemTree` run  and build the content it usually would, then have your if statment in the controller, if its true then pass the CI and the returned treeItemList from the `buildCatalogueItemTree` method and build the response with the ancestor tree. 
if false then return the treeItemList

olliefreeman:
@OButlerOcc can you resolve the issues and see the comment i left please .... its a little awkward as i opened the PR i can't submit a formal PR review....so its all in comments

olliefreeman:
Okay so i misunderstood the ticket...re-read it...please make the following changes as we actually want a clean and new endpoint to ONLY return the ancestors

Endpoint:
get "/${catalogueItemDomainType}/$catalogueItemId/ancestors"(controller: 'treeItem', action: 'ancestors')

Whats in the controller method is right, but copy the controller `show` method ... name the copy `ancestors` and remove the "if" clause you have and just respond with your containerTreeItem, 

then roll back your changes in the show method

copy the show.gson file and call it ancestors.gson and remove the if statements from it and only exoect the containerTreeItem in the model

roll back your changes to the show.gson

olliefreeman (commented):
happy with this

PR #173, opened 2021-10-29
Feature/mc 9510

aaronforshaw:
- Add beans to render ApiProperty or collection of ApiProperty as CSV
- Add .gml files for index and show of ApiProperty, and add MarkupViewConfiguration to MdmCoreGrailsPlugin so that the .gml files are used rather than the default XML renderer
- Add csv to response formats, so when appending ?format=csv or ?format=xml to GET /api/admin/properties or GET /api/admin/properties/{id} the response is rendered as CSV or XML
- Add a data binding bean for CsvDataBindingSourceCreator which enables binding of CSV to either a single target instance or a collection of target instances
- The endpoint POST /api/admin/properties with a body expressed as XML or CSV thus creates a resource in the same way as happens for JSON. The Content-Type header in the POST must be set to application/xml or text/csv
- Add an endpoint POST /api/admin/properties/apply so that a body containing a collection of ApiProperties as JSON, XML or CSV will be "applied" i.e. imported. For XML the Content-Type header on the POST must be set to application/xml. For CSV the Content-Type header on the POST must be set to text/csv. All items in the collection are saved in a single transaction, so a failure in one item (which will happen if the key or value already exists) means that none will be imported.
- Add a CsvComparer trait to make testing easier
- Add tests in mdm-core and mdm-testing-functional

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -23,6 +23,7 @@ hibernateSearchVersion=2.3.0
 luceneVersion=5.5.5
 ## Core
 commonsTextVersion=1.9
+commonsCsvVersion=1.8
 springBootVersion=2.1.17.RELEASE
 jaxbApiVersion=2.3.1
 assetPipelineVersion=3.3.2
--- a/mdm-core/dependencies.gradle
+++ b/mdm-core/dependencies.gradle
@@ -3,6 +3,7 @@ dependencies {
     compile project(':mdm-common')
 
     compile group: 'org.apache.commons', name: 'commons-text', version: commonsTextVersion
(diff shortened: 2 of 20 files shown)


olliefreeman (changes requested):
i dont think the renderer beans should be labelled as HAL ?? surely they should just be csvApiPropertyRenderer

aaronforshaw:
> i dont think the renderer beans should be labelled as HAL ?? surely they should just be csvApiPropertyRenderer

Yes, quite right. I have changed the labels.

olliefreeman (approved):

PR #175, opened 2021-11-02
Add columnNumber to ReferenceDataElement

aaronforshaw:
To ensure that after importing a reference data CSV file the data can be displayed in the same column order as the input file:
- add a zero-based columnNumber property to ReferenceDataElement
- add database migration script which adds this column, sets the value to 0 for all existing values, and then makes the column not null. Ordering is not applied to existing reference data elements
- change CSV importer to set columnNumber as it iterates through the column headers
- update all importers and exporters (CSV, JSON, XML) from assorted versions 3.0, 3.1 to a consistent 4.0 in all cases
- add a version 4.0 xsd for the XML export
- add columnNumber property to various .gson and .gml, including in the exporters
- in cases where it is necessary (API to list ReferenceDataElements and API to list ReferenceDataValues in row format, both of which are consumed by the UI and which I think makes an assumption that results are sorted) ensure that the results are sorted by the new columnNumber attribute. In other places where it seems not necessary, the results are not sorted (but a consumer could choose to sort by the columnNumber property)
- update lots of test data and specs

Also ran some informal tests with the UI to check results are ordered as expected.

### DIFF
--- a/mdm-plugin-referencedata/grails-app/assets/xsd/export_4.0.xsd
+++ b/mdm-plugin-referencedata/grails-app/assets/xsd/export_4.0.xsd
@@ -1,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:exp="http://maurodatamapper.com/export/4.0"
+           xmlns:mdm="http://maurodatamapper.com/referenceDataModel/4.0" elementFormDefault="qualified"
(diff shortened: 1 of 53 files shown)


olliefreeman:
@aaronforshaw you may have missed the messages regarding the weird jenkins behavour there is now a required workflow
1. push branch
2. wait for jenkins to pass that branch
3. open PR

this will prevent the erroneous "branch" build which exist above from being a failure which sometomes happens

olliefreeman (approved):

Issue #171, opened 2021-10-27
Search does not work when user selects a folder in the Advanced options


jelenadreskai:
Folder structure : GENERAL --> L1--> L2, and model is in L2 Folder.
User selects "Search" functionality
In the Advanced options user selects one of the folders 
User enters the search term
Note: enter term* if you want to find any object with prefix "term"

Expected results:
Search results would find the search term in the data models within selected folders, no matter how are folders nested

Actual results:
**Search returns no results if user selects the folder, one or 2 levels up (General or L1)** 
Search returns results as expected, if user selected L2 folder.

** Version ***
UI:  6.6.0 
MDM core: 4.10.0

Question @jamesrwelch  @olliefreeman  @pjmonks : Is this a bug, or done on purpose? 

Gail does have a fix, but wondering whether this needs to be configurable? So the admin can allow nested folder searches or not. I do not see a reason why would it, since the MDM does offer a search in all folders by default.

jelenadreskai:
@olliefreeman can you assign this to Gail please.

PR #190, opened 2021-11-16
GH-171 Fix for folder search to include nested folders

GBishop-PHSA:


olliefreeman (approved):
I'd prefer it if the base folder which all the test use is not a sub folder.
It just means we dont need to worry about hierarchy during testing some of the more basic concepts. However i guess this setup should highlight any issues earlier

olliefreeman:
@GBishop-PHSA  can you add a brief description to the PR in the future. And also make sure any issues the PR resolves are also linked. We're using the GH built in doc generation for release notes and it uses issues and PR descriptions and names to provide info

PR #198, opened 2021-11-24
Breadcrumb not set updated to check for is finalised and set breadcru…

gammonpeter:
…mbs as required


olliefreeman (changes requested):
i assume this works, would like a test to show it.
 i would also like to know why the BTs arent set correctly during the actual import as the DM isnt saved until all the stuff is done, by which point the DM is finalised so the DM contents should know the DM is finalised and therefore the BTs should be correct

PR #200, opened 2021-11-25
Feature/gh 182

### LINKED ISSUE #182: Copy class fails with validation error
**Describe the bug**
When copying a data class that has other properties on the Data Class and other properties on the Data Element, the newly created class fails to save with a validation error on the Data Element property.

The failure happens at: https://github.com/MauroDataMapper/mdm-core/blob/09bc29cc2bd703a1c304ff9b4ff8fb681e89a068/mdm-core/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/core/facet/Metadata.groovy#L46

because ident() returns null.

**To Reproduce**
1. Create a new Data Model
2. Create a new Data Class for the data model
3. Go to Other Properties and add a property to the data class
4. Create a new Data Element for the data class
5. Go to Other Properties for the Data Element
6. Go to a different Data Model than the one created above
7. Create a new Data Class by using Copy Data Class(es) from
8. Select the original Data Modal and click Next Step
9. Select the Data Class created above and click Submit Data Class button
10. See error: - Field error in object 'uk.ac.ox.softeng.maurodatamapper.core.facet.Metadata' on field 'multiFacetAwareItemId': rejected value [null];

**Expected behavior**
The new copied data class should be created with Other Properties on the class and data element level created.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Chrome
 - Version mdm-core 4.10.0

**Additional context**
When debugging, the validation error occurs when saving the Data Element other property.

GBishop-PHSA:
Fixes #182

### DIFF
--- a/mdm-plugin-datamodel/grails-app/services/uk/ac/ox/softeng/maurodatamapper/datamodel/item/DataClassService.groovy
+++ b/mdm-plugin-datamodel/grails-app/services/uk/ac/ox/softeng/maurodatamapper/datamodel/item/DataClassService.groovy
@@ -79,6 +79,7 @@ class DataClassService extends ModelItemService<DataClass> implements SummaryMet
     @Override
     DataClass save(Map args, DataClass domain) {
         // If not previously saved then allow a deep save and/or datatype save
+        Collection<DataType> newDataTypes
         if (!domain.ident()) {
             if (args.deepSave) {
                 saveDataClassHierarchy(domain)
@@ -86,13 +87,37 @@ class DataClassService extends ModelItemService<DataClass> implements SummaryMet
 
             // If ignore datatypes then skip this bit or DC already been saved as this is designed to handle full builds or copies
             if (args.saveDataTypes) {
-                saveDataTypesUsedInDataClass(domain)
+                newDataTypes = saveDataTypesUsedInDataClass(domain)
             }
         }
-        super.save(args, domain)
+        if(args.deepSave){
+            this.updateDataClassHierarchyAfterInsert(args, domain)
+        }
+
+        DataClass saved = super.save(args, domain)
+
+        if(newDataTypes){
+            newDataTypes.each{dt ->
+                dataTypeService.updateFacetsAfterInsertingCatalogueItem(dt)
+            }
+        }
+        saved
     }
 
-    void saveDataTypesUsedInDataClass(DataClass dataClass) {
+    void updateDataClassHierarchyAfterInsert(Map args, DataClass dataClass){
+        if(dataClass.dataElements){
+            dataClass.dataElements.each{de ->
+                dataElementService.save(de)
+            }
+        }
+        if(dataClass.dataClasses){
+            dataClass.dataClasses.each{dc ->
+                updateDataClassHierarchyAfterInsert(args, dc)
+            }
+        }
+    }
+
+    Collection<DataType> saveDataTypesUsedInDataClass(DataClass dataClass) {
         // Make sure all datatypes are saved
         Set<DataType> dataTypes = extractAllUsedNewOrDirtyDataTypes(dataClass)
         log.debug('{} new or dirty used datatypes inside dataclass', dataTypes.size())
@@ -123,7 +148,6 @@ class DataClassService extends ModelItemService<DataClass> implements SummaryMet
         dataClass.dataElements?.clear()
         Set<ReferenceType> referenceTypes = new HashSet<>(dataClass.referenceTypes ?: [])
         dataClass.referenceTypes?.clear()
-
         // Save the DC without flushing
         dataClass.save(flush: false, validate: false)
 
--- a/mdm-plugin-datamodel/grails-app/utils/uk/ac/ox/softeng/maurodatamapper/datamodel/bootstrap/BootstrapModels.groovy
+++ b/mdm-plugin-datamodel/grails-app/utils/uk/ac/ox/softeng/maurodatamapper/datamodel/bootstrap/BootstrapModels.groovy
@@ -147,6 +147,7 @@ class BootstrapModels {
 
             DataClass parent =
                 new DataClass(createdBy: DEVELOPMENT, label: 'parent', minMultiplicity: 1, maxMultiplicity: -1, dataModel: dataModel)
+
             DataClass child = new DataClass(createdBy: DEVELOPMENT, label: 'child')
             parent.addToDataClasses(child)
             dataModel.addToDataClasses(child)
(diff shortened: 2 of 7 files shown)


GBishop-PHSA:
This pull request is an initial attempt at fixing the Class Copy issue when a Class has metadata and at least one of its Elements has metadata. The metadata is not saved after the copy, and the validation fails for the parent catalogue item id.

Change to the BootstrapModels was to add metadata at the class and element level for the Class: 'parent'. This causes the test 'C01 : test copying (as writer to both DataModels)' in uk.ac.ox.softeng.maurodatamapper.testing.functional.datamodel.item.DataClassFunctionalSpec to fail with  the validation error.

mdm-plugin-datamodel/grails-app/utils/uk/ac/ox/softeng/maurodatamapper/datamodel/bootstrap/BootstrapModels.groovy
```diff
@@ -175,6 +185,15 @@ class BootstrapModels {
             refType.addToDataElements(el1)
             parent.addToDataElements(el1)
 
+            checkAndSave(messageSource, parent)
+            Metadata meta = new Metadata(createdBy: DEVELOPMENT,
+                                         namespace: 'test.com', key: 'mdel1', value: 'mdel1',
+                                         multiFacetAwareItemId: el1.id,
```
olliefreeman, line 191: if you look at the`uk.ac.ox.softeng.maurodatamapper.core.traits.domain.MultiFacetItemAware.setMultiFacetAwareItem` which is whats used during "map" constructors the id will be set when you set the MFAI, also the addToMetadata will set the MFAI as well, so the cleaner way is to create the metadata without those 2 fields, then addToMetadata then checkAndSave the MFAI then the MD

mdm-plugin-datamodel/grails-app/services/uk/ac/ox/softeng/maurodatamapper/datamodel/item/DataClassService.groovy
```diff
@@ -203,6 +205,18 @@ class DataClassService extends ModelItemService<DataClass> implements SummaryMet
     @Override
     DataClass updateFacetsAfterInsertingCatalogueItem(DataClass catalogueItem) {
         super.updateFacetsAfterInsertingCatalogueItem(catalogueItem)
+
```
olliefreeman, line 208: this is not right, it doesnt properly handle several issues and also will then be called on all DC saves which isnt appropriate.

olliefreeman (changes requested):
So stepping through the copy->save

we do 
1. `uk.ac.ox.softeng.maurodatamapper.core.controller.CatalogueItemController#saveResource`
2. `uk.ac.ox.softeng.maurodatamapper.core.controller.CatalogueItemController#serviceInsertResource`
3. `uk.ac.ox.softeng.maurodatamapper.datamodel.item.DataClassController#serviceInsertResource`
4. `uk.ac.ox.softeng.maurodatamapper.datamodel.item.DataClassService#save(java.util.Map, uk.ac.ox.softeng.maurodatamapper.datamodel.item.DataClass)`

at this point deepSave == true and saveDataTypes == true.

we cant save DEs until its parent DC and its DT is saved, which is why the `saveDataClassHierarchy` removes all content from a DC and then we save the DTs and then the standard `uk.ac.ox.softeng.maurodatamapper.core.model.CatalogueItemService#save(java.util.Map, K)` method is called as this cascades the save to the DEs

The appropriate way to do this is to accept that this method already has custom saving in place so we should save the result of `super.save(args, domain)` and then do some post processing

```groovy
    @Override
    DataClass save(Map args, DataClass domain) {
        // If not previously saved then allow a deep save and/or datatype save
        Collection<DataTypes> newDataTypes
        if (!domain.ident()) {
            if (args.deepSave) {
                saveDataClassHierarchy(domain)
            }

            // If ignore datatypes then skip this bit or DC already been saved as this is designed to handle full builds or copies
            if (args.saveDataTypes) {
                newDataTypes = saveDataTypesUsedInDataClass(domain)
            }
        }

        DataClass saved = super.save(args, domain)
        if(args.deepSave){
            updateDataClassHierarchyAfterInsert(saved)
        }
        if(newDataTypes){
        newDataTypes.each{dt ->
            dataTypeService.updateFacetsAfterInsertingCatalogueItem(dt)
        }
        }
      saved
  }

    Collection<DataTypes> saveDataTypesUsedInDataClass(DataClass dataClass) {
    the last line of this is a saveAll which returns a list, so just need to update the method to return
}

void updateDataClassHierarchyAfterInsert(DataClass dataClass){
    if(dataClass.dataElements){
    dataClass.dataElements.each{de ->
    dataElementService.updateFacetsAfterInsertingCatalogueItem(de)
    }
    }
    
    if(dataClass.dataClasses){
    dataClass.dataClasses.each{dc ->
    updateDataClassHierarchyAfterInsert(dc)
    }
    
    }
}
```

hopefully this should help

GBishop-PHSA:
Ready for re-review. I've implemented the changes to the Bootstrap data setup and also to the DataClassService. I made an addition, where the DataElements are saved without validation before the Data Class save with validation is called.

mdm-plugin-datamodel/grails-app/services/uk/ac/ox/softeng/maurodatamapper/datamodel/item/DataClassService.groovy
```diff
@@ -79,20 +79,46 @@ class DataClassService extends ModelItemService<DataClass> implements SummaryMet
...
+            newDataTypes.each{dt ->
+                dataTypeService.updateFacetsAfterInsertingCatalogueItem(dt)
+            }
+        }
+        saved
     }
 
-    void saveDataTypesUsedInDataClass(DataClass dataClass) {
+    void updateDataClassHierarchyAfterInsert(Map args, DataClass dataClass){
+        if(dataClass.dataElements){
```
olliefreeman, line 109: if you call dataElementService.save() here it will do the save and the update of the facets all in one go as this method is the preferred way to save anything as it passes up to the catalogueItemService.save method

mdm-plugin-datamodel/grails-app/services/uk/ac/ox/softeng/maurodatamapper/datamodel/item/DataClassService.groovy
```diff
@@ -79,20 +79,46 @@ class DataClassService extends ModelItemService<DataClass> implements SummaryMet
...
 
             // If ignore datatypes then skip this bit or DC already been saved as this is designed to handle full builds or copies
             if (args.saveDataTypes) {
-                saveDataTypesUsedInDataClass(domain)
+                newDataTypes = saveDataTypesUsedInDataClass(domain)
             }
         }
-        super.save(args, domain)
+        if(args.deepSave){
+            this.saveDataElements(domain)
```
olliefreeman, line 94: not needed if you make the change in the updateDCHAfterInsert

mdm-plugin-datamodel/grails-app/services/uk/ac/ox/softeng/maurodatamapper/datamodel/item/DataClassService.groovy
```diff
@@ -107,6 +133,18 @@ class DataClassService extends ModelItemService<DataClass> implements SummaryMet
         dataTypes
     }
 
+    void saveDataElements(DataClass dataClass) {
```
olliefreeman, line 136: Can remove this whole method

olliefreeman (changes requested):
All the ModelServices and ModelItemServices extend the CatalogueItemService which handles all the facet fixing as part of the save method. If you always make sure you use the relevant service method to save anything you should find that all issues are taken care of. 

Your fix here is to make sure the DCService handles addtl issues when DCs are saved. Thereby making the Service save method even more reliable

GBishop-PHSA:
@olliefreeman I've removed the unneeded saveElements method and called the service save method. Ready for re-review.

olliefreeman (changes requested):
There are a number of failing tests which need to be fixed, however i'm happy with the code

olliefreeman (approved):

PR #201, opened 2021-11-26
Subscribed models newer versions

joe-crawford:
- Add newerVersions endpoint for published/subscribed models on local and remote catalogues (mc-9726)
- Fix to version link importing for subscribed models (mc-9733) and subscribed model security
- Remove subscribed model federate method and make federate always part of save (mc-9724)
- Separate model label and version in published models list (mc-9729)
- Extend federation client timeout


joe-crawford:
Have marked SubscribedModelFunctionalSpec test as ignore because these connect to the Continuous Deployment instance. However they can successfully from Jenkins.

Also N.B. `V2_3_0__cleanup_subscribedmodel_securableresourcegrouproles.sql` in mdm-security contains a reference to a table in mdm-plugin-federation.

mdm-security/grails-app/conf/db/migration/security/V2_3_0__cleanup_subscribedmodel_securableresourcegrouproles.sql
```diff
@@ -0,0 +1,5 @@
+-- Delete orphaned securable resource group roles for non-existent subscribed models
+DELETE
+FROM security.securable_resource_group_role
+WHERE securable_resource_domain_type = 'SubscribedModel' AND
+      securable_resource_id NOT IN (SELECT id FROM federation.subscribed_model);
```
olliefreeman, line 5: You need to make this work when the federation schema and this table does not exist, we cant guarantee that the security will run before federation or that the federation module hasnt been included

mdm-plugin-federation/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/federation/PublishedModel.groovy
```diff
@@ -20,9 +20,11 @@ package uk.ac.ox.softeng.maurodatamapper.federation
 import uk.ac.ox.softeng.maurodatamapper.core.model.Model
 import uk.ac.ox.softeng.maurodatamapper.version.Version
 
-import java.time.OffsetDateTime
+import groovy.transform.Sortable
 
+import java.time.OffsetDateTime
 
+@Sortable(includes = ['modelLabel', 'modelType', 'modelVersion'])
```
olliefreeman, line 27: When we build the tree (which is the only place we really do ordering of multiple models) we dont really care about modelType. I feel that either we do modelType first in ordering or we remove it

mdm-plugin-federation/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/federation/web/FederationClient.groovy
```diff
@@ -92,7 +92,7 @@ class FederationClient {
         } else {
             this.contextPath = 'api'
         }
-        httpClientConfiguration.setReadTimeout(Duration.ofMinutes(5))
+        httpClientConfiguration.setReadTimeout(Duration.ofMinutes(30))
```
olliefreeman, line 95: 30mins for a live connection to a server is very very long, what you're saying here is hold at this point in the code for 30mins for a response, this could collapse the UI as the UI will wait for a response from us before renderering the tree. 
I think that its acceptable to make this a configurable option so that if a server is slow then the sys admins can up the timeout but 30mins should not be hardcoded.
I think the best option would be a configurable timeout for the SubscribedCatalogue so that each SC can be customised, and then a default timeout in the yaml files using grailsApplication.config.<> to acess it. The prop should be `maurodatamapper.federation.subscribedcatalogue.default.timeout: 5`

olliefreeman (changes requested):
Happy with the vast majority of this, looks good, just need to make a couple of changes.
* SQL migration needs to handle the federation not existing
* Configurable timeout on SC with a default of 5m not 30
* Sort ordering wrt ModelType

mdm-security/grails-app/conf/db/migration/security/V2_3_0__cleanup_subscribedmodel_securableresourcegrouproles.sql
joe-crawford: Have added a DO block with a check that the table exists.

olliefreeman (approved):

PR #217, opened 2021-12-13
Feature/gh 209 - fix profile sort

aaronforshaw:
- Fix the way profiles are sorted, so that when a profile having two versions, each with version tags, is retrieved by namespace and name only, the most recent semantic version is retrieved.
- Exclude non-finalised dynamic profiles from the profiles listed at the ```/{multiFacetAwareItemDomainType}/{multiFacetAwareItemId}/profiles/used``` and ```/{multiFacetAwareItemDomainType}/{multiFacetAwareItemId}/profiles/unused``` endpoints


olliefreeman (changes requested):
Can you also add a test and update the code to only return finalised versions in the used/unused endpoints. This will help the UI fix #402, by at least not showing branches

olliefreeman (approved):

PR #223, opened 2021-12-22
Feature/mc 9631 - Branch and merge Model Data Type

aaronforshaw:
- Fix bugs in the path retrieval of VersionedFolder.
- Add a method to the PathService to find a resource by its Path only, with no security check on the securable resource which owns the resource.
- When branching a Versioned Folder, update Model Data Types in the branched folder so that they point to the branched model.
- Update ModelDataType.diff so that it only returns a diff if the paths of the pointed to resource in source and target are different after ignoring differences in the branch and version. The diff returned is on 'modelResourcePath' rather than modelResourceId and modelResourceDomainType. This 'modelResourcePath' is computed as a fully qualified path by recursing up all parents of the model.
- Add a method to DomainService which enables subclasses to state that they will provide special handling for the patching a modification into a Versioned Folder
- In ModelDataTypeService, override the above method and use the 'modelResourcePath' provided in the diff to look up the relevant resource to point to when merging.
- Ditto the above two points, but for merging when not in a Versioned Folder
- Add test data
- Add tests for (i) branching of Model Data Type in a VF (ii) merge a Model Data Type when it is pointing to a resource in the source folder (iii) merge a Model Data Type when it is pointing to a resource outside of the VF
- Add test for branching and merging of a Data Model containing a Model Data Type when not in a VF

### DIFF
--- a/mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/model/ModelService.groovy
+++ b/mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/model/ModelService.groovy
@@ -1076,9 +1076,8 @@ abstract class ModelService<K extends Model>
      *
      * @param copiedModel
      * @param originalModel
-     * @param branchName The name of the versioned folder branch to which copiedModel has been copied
      */
-    void updateCopiedCrossModelLinks(K copiedModel, K originalModel, String branchName) {
+    void updateCopiedCrossModelLinks(K copiedModel, K originalModel) {
         log.debug('Updating cross model links for [{}]', Path.from(copiedModel))
 
         // TODO
--- a/mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/model/ModelService.groovy
+++ b/mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/model/ModelService.groovy
@@ -599,9 +599,14 @@ abstract class ModelService<K extends Model>
         }
         String fieldName = modificationPatch.fieldName
         log.debug('Modifying [{}] in [{}]', fieldName, modificationPatch)
-        domain."${fieldName}" = modificationPatch.sourceValue
         DomainService domainService = getDomainServices().find {it.handles(domain.class)}
         if (!domainService) throw new ApiInternalException('MSXX', "No domain service to handle modification of [${domain.domainType}]")
+
+        // If the domainService provides a special handler for modifying this field then use it,
+        // otherwise just set the value directly
+        if (!domainService.handlesModificationPatchOfField(modificationPatch, domain, fieldName)) {
+            domain."${fieldName}" = modificationPatch.sourceValue
+        }
 
         if (!domain.validate())
             throw new ApiInvalidModelException('MS01', 'Modified domain is invalid', domain.errors, messageSource)
--- a/mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/model/ModelService.groovy
+++ b/mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/model/ModelService.groovy
@@ -604,7 +604,7 @@ abstract class ModelService<K extends Model>
 
         // If the domainService provides a special handler for modifying this field then use it,
         // otherwise just set the value directly
-        if (!domainService.handlesModificationPatchOfField(modificationPatch, domain, fieldName)) {
+        if (!domainService.handlesModificationPatchOfField(modificationPatch, targetModel, domain, fieldName)) {
             domain."${fieldName}" = modificationPatch.sourceValue
         }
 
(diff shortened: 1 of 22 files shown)


mdm-plugin-datamodel/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/datamodel/item/datatype/ModelDataType.groovy
```diff
@@ -41,10 +46,34 @@ class ModelDataType extends DataType<ModelDataType> {
     }
 
     ObjectDiff<ModelDataType> diff(ModelDataType otherDataType, String context) {
-        catalogueItemDiffBuilder(ModelDataType, this, otherDataType)
-            .appendString('modelResourceId', this.modelResourceId.toString(), otherDataType.modelResourceId.toString())
-            .appendString('modelResourceDomainType', this.modelResourceDomainType, otherDataType.modelResourceDomainType)
+        ObjectDiff<ModelDataType> diff = catalogueItemDiffBuilder(ModelDataType, this, otherDataType)
 
+        GrailsApplication grailsApplication = Holders.getGrailsApplication()
```
olliefreeman, line 51: i'd prefer it if we used a transient field `model` which we then populated inside one of the services with the models, rather than doing domain lookups inside a domain.
We have a similar transient field in the Facet class.
 Also you've loaded these models already in the DMS, so you can just push the loaded model into the transient field

mdm-plugin-datamodel/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/datamodel/item/datatype/ModelDataType.groovy
```diff
@@ -41,10 +46,34 @@ class ModelDataType extends DataType<ModelDataType> {
...
+            List<Model> otherResourceModels = otherResourceClass.byIdInList([otherDataType.modelResourceId]).list()
+
+            if (thisResourceModels.size() == 1 && otherResourceModels.size() == 1) {
+                Model thisResourceModel = thisResourceModels.first()
+                Model otherResourceModel = otherResourceModels.first()
+
+                // Aside from branch and version, is the model pointed to by the modelDataType really different by path?
+                Path thisResourcePath = Path.from(thisResourceModel.folder, thisResourceModel)
+                Path otherResourcePath = Path.from(otherResourceModel.folder, otherResourceModel)
+                if (!thisResourcePath.equalsIgnoringModelIdentifier(otherResourcePath)) {
```
olliefreeman, line 67: you can use matches() here

mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/model/ModelService.groovy
```diff
@@ -1038,7 +1038,13 @@ abstract class ModelService<K extends Model>
...
     }
 
-    void updateCopiedCrossModelLinks(K copiedModel, K originalModel) {
+    /**
+     *
+     * @param copiedModel
+     * @param originalModel
+     * @param branchName The name of the versioned folder branch to which copiedModel has been copied
+     */
+    void updateCopiedCrossModelLinks(K copiedModel, K originalModel, String branchName) {
```
olliefreeman, line 1047: Dont see why you need the branch name, you have the copied and original model which means you have the branch name as its the copiedModel.branchName field

olliefreeman (changes requested):
Looks like the methodology is right, 

All the stuff in path and path node and model, can you replace it all with the use of  `boolean matchesIdentifier(PathNode otherPathNode, String modelIdentifierOverride = null)` as thats why that method exists, you can pass in a branch name or model version number as the override and it will do everything you've coded up for you.

mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/model/ModelService.groovy
aaronforshaw: Done

mdm-plugin-datamodel/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/datamodel/item/datatype/ModelDataType.groovy
aaronforshaw: Done

mdm-plugin-datamodel/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/datamodel/item/datatype/ModelDataType.groovy
aaronforshaw: I have not done this. In the DMS the models are available in DataModelService.updateCopiedCrossModelLinks. This method is used on branching, whereas in ModelDataType the models are required when merging.

aaronforshaw:
@olliefreeman Please see updated description at the top of this PR.

olliefreeman (approved):
with the changes in commit im happy with this now

olliefreeman:
failing test is a known flaky test so merging

PR #229, opened 2022-01-06
Feature/mc 9632 - Branch and merge Reference Data Models

aaronforshaw:
Make branching and merging work for Reference Data Models.

- Fix various pathing bugs for the items of Reference Data Model
- Fix copying of Reference Data Element and Reference Data Type
- Add a ReferenceDataPluginMergeBuilder to create test data
- Copy tests from DataModelFunctionalSpec to ReferenceDataModelFunctionalSpec, with suitable adaptations
- Update the VersionedFolderMergeBuilder and VersionedFolderFunctionalSpec to include Reference Data
- Use a special handler to deal with referenceDataTypePath, with the same pattern as for Model Data Type


mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/container/VersionedFolderService.groovy
```diff
@@ -871,7 +872,7 @@ class VersionedFolderService extends ContainerService<VersionedFolder> implement
         }
         String fieldName = modificationPatch.fieldName
         log.debug('Modifying [{}] in [{}]', fieldName, modificationPatch.path.toString(getModelIdentifier(targetVersionedFolder)))
-        domain."${fieldName}" = modificationPatch.sourceValue
+        CreatorAwareUtils.patchCreatorAwareField(domain, fieldName, modificationPatch.sourceValue)
```
olliefreeman, line 875: see the main review...this shouldnt be happening

mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/model/ModelService.groovy
```diff
@@ -565,7 +566,7 @@ abstract class ModelService<K extends Model>
         }
         String fieldName = modificationPatch.fieldName
         log.debug('Modifying [{}] in [{}]', fieldName, modificationPatch)
-        domain."${fieldName}" = modificationPatch.sourceValue
+        CreatorAwareUtils.patchCreatorAwareField(domain, fieldName, modificationPatch.sourceValue)
```
olliefreeman, line 569: see main review

olliefreeman (changes requested):
Apart from the below i'm happy with this.

I disagree with the need for the CreatorAwareUtils:
 Diffs should not be detecting the change at the subdomain level. i.e. `domain.dataType.label` should not be a detected change of domain. It should be a detected change of dataType, the domain should not care that the dataType label has changed. If the datatype has changed then the id would be different, otherwise the "pointer" would remain the same and the diff would state that the datatype itself has been modified.
 
 Can you provide the exact situation this is happening and why its needed, for me to better understand

aaronforshaw:
> Apart from the below i'm happy with this.
> 
> I disagree with the need for the CreatorAwareUtils: Diffs should not be detecting the change at the subdomain level. i.e. `domain.dataType.label` should not be a detected change of domain. It should be a detected change of dataType, the domain should not care that the dataType label has changed. If the datatype has changed then the id would be different, otherwise the "pointer" would remain the same and the diff would state that the datatype itself has been modified.
> 
> Can you provide the exact situation this is happening and why its needed, for me to better understand

@olliefreeman At

https://github.com/MauroDataMapper/mdm-core/blob/e202601fc265d2a49ae0248310f312f16b929513/mdm-plugin-datamodel/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/datamodel/item/DataElement.groovy#L198-L205 

there is a string appended with dataType.label

Similarly in ReferenceDataElement (which is presumably copied from DataElement).

olliefreeman:
as discussed in the meeting, we need to change this to "path" for the datatype comparison not label, then use the same coding format being developed for the model datatypes

aaronforshaw:
Updated to use the pathing approach for referenceDataTypeLabel, and also changed the tests following the removal of isLegacy option on endpoints.
The test failures appear to be inconsistent and not related to this change.

PR #231, opened 2022-01-10
Close federation connections and published models version number fix

joe-crawford:


olliefreeman (approved):
Happy with this, we will need to remember the whole closeable thing with anything else that we do.

PR #252, opened 2022-02-11
Feature/gh 247 - Disallow duplicate keys on apply; set lastUpdatedBy on update

### LINKED ISSUE #247: Creating API properties collection allows duplicate keys
**Describe the bug**
API property keys should be unique. When creating keys via a collection, duplicate keys are accepted.

**To Reproduce**
Using Postman, repeatedly post the following to `/admin/properties/apply`:
```xml
<?xml version='1.0'?>
<apiProperties>
    <count>2</count>
    <items>
        <apiProperty>
            <key>a.first.xml.key</key>
            <value>A first XML value</value>
            <category>Email</category>
            <publiclyVisible>false</publiclyVisible>
        </apiProperty>
        <apiProperty>
            <key>a.second.xml.key</key>
            <value>A second XML value</value>
            <category>Email</category>
            <publiclyVisible>false</publiclyVisible>
        </apiProperty>
    </items>
</apiProperties>
```
When listing the API properties (`/admin/properties`), multiple instances of `a.first.xml.key` and `a.second.xml.key` occur.

**Expected behavior**
The API properties collection creation endpoint `/admin/properties/apply` should reject collections which contain duplicate keys.

**Additional context**
Tested on latest `develop`.

aaronforshaw:
Closes #247 

- Change the endpoint ```POST /admin/properties/apply``` so that each instance of the posted collection is validated before save. If any item in the collection fails validation then reject the entire collection. 
- Add tests to cover both a replay of the same collection, and also duplicate keys in the same collection
- Change the endpoint ```PUT /admin/properties/{id}``` to set the lastUpdatedBy field. Add a test for this.

### DIFF
--- a/mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/admin/ApiPropertyController.groovy
+++ b/mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/admin/ApiPropertyController.groovy
@@ -67,6 +67,16 @@ class ApiPropertyController extends EditLoggingController<ApiProperty> {
         saveResponse instance
     }
 
+    /**
+     * Override the base method so that we can set instance.lastUpdatedBy
+     */
+    @Transactional
+    @Override
+    protected boolean validateResource(ApiProperty instance, String view) {
+        instance.lastUpdatedBy = currentUser.emailAddress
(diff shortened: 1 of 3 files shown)


olliefreeman (approved):

aaronforshaw:
I think the test failure is not related to this change

PR #263, opened 2022-02-15
Feature/gh 222

### LINKED ISSUE #222: Cannot merge a patch which includes a change of DT for a DE
Found inside the NHSDD, it would appear the label of the DT changed, i suspect this is actually that the DT itself changed.
However the patch below could not be applied because `groovy.lang.MissingPropertyException: No such property: dataType.label for class: uk.ac.ox.softeng.maurodatamapper.datamodel.item.DataElement`

We need to work out how to handle the DT changing for a DE inside a branch, as if we only check the DT itself then every single DE will have a merge modification as all the ids will have changed. However we do need to identify that the DT has changed and create an appropriate patch/diff.

```json
    {
      "fieldName": "dataType.label",
      "path": "vf:NHS Data Dictionary (September 2021)$main|dm:NHS Data Dictionary - Core$main|dc:Attributes|de:Admission Source@dataType.label",
      "sourceValue": "Admission Source Attribute Type",
      "targetValue": "String",
      "commonAncestorValue": "String",
      "isMergeConflict": false,
      "type": "modification"
    },
```

We can see the issue with only looking at IDs in the ModelDataType where the `modelResourceId` field is always flagged as having chnaged between 2 branches. This is also probably something to be solved in this ticket

aaronforshaw:
Closes #222 

Handle merging when the Data Type of a Data Element has been changed in a branch.

- Use ```dataTypePath``` rather than ```dataType.label``` as a diff identifier, and use pathing to identify when the Data Type of a Data Element has really ben changed
- Update merge test data builder and test results

### DIFF
--- a/mdm-plugin-datamodel/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/datamodel/item/DataElement.groovy
+++ b/mdm-plugin-datamodel/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/datamodel/item/DataElement.groovy
@@ -184,12 +184,19 @@ class DataElement implements ModelItem<DataElement, DataModel>, MultiplicityAwar
     }
 
     ObjectDiff<DataElement> diff(DataElement otherDataElement, String context) {
-        catalogueItemDiffBuilder(DataElement, this, otherDataElement)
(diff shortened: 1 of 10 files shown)


aaronforshaw:
I think the test failure on the branch is not related to this change.

olliefreeman (approved):

PR #280, opened 2022-02-28
Copy index order with ModelItems

### LINKED ISSUE #241: Data Class ordering not maintained when creating a new version
When creating a new version of a model, the classes are copied but the ordering is not maintained.  This happened within a versioned folder, but I assume the issue applies on a single model as well.  The `idx` field in the database is set, but not with the correct ordering.

joe-crawford:
- When copying (reference) data classes, elements, types and enumeration values as part of another item, copy their indices (resolves #241)

### DIFF
--- a/mdm-plugin-terminology/grails-app/services/uk/ac/ox/softeng/maurodatamapper/terminology/item/TermService.groovy
+++ b/mdm-plugin-terminology/grails-app/services/uk/ac/ox/softeng/maurodatamapper/terminology/item/TermService.groovy
@@ -236,12 +236,19 @@ class TermService extends ModelItemService<Term> {
         copy
     }
 
+    Term copyTerm(CodeSet copiedCodeSet, Term original, User copier, UserSecurityPolicyManager userSecurityPolicyManager,
+                  CopyInformation copyInformation = null) {
+        Term copy = copyTerm(original, copier, userSecurityPolicyManager, copyInformation)
+        copiedCodeSet.addToTerms(copy)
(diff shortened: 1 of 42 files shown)


mdm-plugin-terminology/grails-app/services/uk/ac/ox/softeng/maurodatamapper/terminology/item/TermService.groovy
```diff
@@ -236,6 +236,13 @@ class TermService extends ModelItemService<Term> {
         copy
     }
 
+    Term copyTerm(CodeSet copiedCodeSet, Term original, User copier, UserSecurityPolicyManager userSecurityPolicyManager,
```
olliefreeman, line 239: so close to not having any comments....

we never copy terms when we copy a codeset as terms are owned by a terminology, when copying a codeset we always use the existing term and just add the existing term to the copied codeset, therefore this method needs to be removed

olliefreeman (changes requested):
Only 1 issue that needs fixing

joe-crawford:
Failing test looks to be unrelated

mdm-plugin-terminology/grails-app/services/uk/ac/ox/softeng/maurodatamapper/terminology/item/TermService.groovy
joe-crawford: Have removed.

PR #284, opened 2022-03-04
mc-9808 Allow a simplified creation for ReferenceTypes and ModelDataTypes

olliefreeman:
* Allow POST to DataTypes without a label for RTs and MDTs
* Generate a default label of "Reference to X" where X is the label of the reference
* Confirm the existence of the reference
* If a DT already exists with that default label then generate a new one with the next number in order suffixed
* Allow POST to DataElements without a label in the DataType for creating new DTs alongside the DE
* Use the default label for RTs and MDTs when creating a DE if no label supplied
* Try and find existing DT with same domaintype and label when creating a DE and DT at the same time.
 If the DT exists then use that one rather than creating a new one which will fail as the DT already exists


olliefreeman:
test failures are to do with timings, not the code. need to increase the timeout for lock.

PR #291, opened 2022-03-11
mc-9827 add method to send email with any properties map

aaronforshaw:
Adds a method to send an email to a user. The properties map is provided by the caller of this method, so that an external plugin can specify its own email body and its own properties map.


aaronforshaw:
I think the test failures on the pr-merge are not related to this change

olliefreeman (approved):

Issue #295, opened 2022-03-17
Internal Server Error when subsetting many data elements


aaronforshaw:
# Description

Using the subset endpoint with many additions can cause an exception.

# Steps to reproduce

1. Create a source data model containing a class with lots of data elements 
2. Create an empty target data model
3. Using postman, construct a a request to 
```
PUT {{base_url}}/dataModels/{sourceDataModelId}/subset/{targetDataModelId}
```
with a body containing, for example, 
```
{"additions":["fbfb1a53-42d2-4437-b601-1a90d667f7c0", "6d3a0d42-93e5-427a-8203-a38c2f7a5fd0", "43dbed3c-e1f8-49b1-9ba6-43c3dfa977cc", "bdf06076-aa8e-49e2-97e8-f86fcb521ab9", "27e87165-4acb-4efa-bf87-898954c87e68", "ea8c92e7-a308-44b4-983d-3e053e86caff", "fe7ecaaa-cf86-4e26-8e1d-6db030cd9b56", "6eb3975a-c932-4ae5-8f84-83f15b49f375", "25e52261-f6b1-4e29-bf2b-f35997e9feac", "56bd2761-5a8e-42ca-b7d0-fb4e843cb4ef"], "deletions": []}
```
# Expected behavior
All data elements are saved to the target data model

# Screenshots
```
{
    "status": 500,
    "reason": "Internal Server Error",
    "errorCode": "UEX--",
    "message": "Batch update returned unexpected row count from update [0]; actual row count: 0; expected: 1; statement executed: update datamodel.data_type set version=?, date_created=?, domain_type=?, last_updated=?, path=?, breadcrumb_tree_id=?, data_model_id=?, idx=?, created_by=?, aliases_string=?, label=?, description=?, units=? where id=? and version=?",
    "path": "/api/dataModels/8d9ada0f-a53c-470d-b947-cc7a7f80dad5/subset/c9a84242-1d4a-4068-9593-6415baee64f6",
    "environment": "DEVELOPMENT",
    "version": "5.1.0-SNAPSHOT",
    "exception": {
        "type": "StaleStateException",
        "message": "Batch update returned unexpected row count from update [0]; actual row count: 0; expected: 1; statement executed: update datamodel.data_type set version=?, date_created=?, domain_type=?, last_updated=?, path=?, breadcrumb_tree_id=?, data_model_id=?, idx=?, created_by=?, aliases_string=?, label=?, description=?, units=? where id=? and version=?",
[...]

PR #296, opened 2022-03-17
gh-295 flush when saving data class and data element in the subset operation

aaronforshaw:
Fix a "StaleStateException" by flushing each save of target class and data element


aaronforshaw:
@olliefreeman I don't really understand why this fixes the problem (which was perhaps timing related), but it seems to

olliefreeman:
fixes #295

olliefreeman:
given the error message it should only be the DE which needs to be flushed  .. `actual row count: 0; expected: 1; statement executed: update datamodel.data_type`. Whats happening is the DT is being created/updated for the inclusion/exclusion of a DE, the DE and DC are being saved into the persistence context however the DT is not, its being "loosely" held by the DE. Flushing the DE causes the system to identify a change to the DT and therefore save it, however in the DMController when the save is called the DT change is not being detected (ORM being great here). This is why flushing on the DE works.

There shouldnt be an issue with this going forwards, however if anythign else arises the DMService "saveModelWithNewContentOnly" will need to be used rather than the default updateResource in the DMController, and flushes will need to be removed from the subset method.

PR #309, opened 2022-05-23
Subscribed catalogue support for Atom XML and importing from published URLs

joe-crawford:
Add support for subscribing to Atom XML feeds as subscribed catalogues, and to the Mauro JSON format published in PR #309:

- Add type field to Subscribed Catalogue (`Mauro JSON` or `Atom`) and for Atom feeds, store the whole URL to the Atom XML feed. Convert from JSON or Atom formats to published model list in the local catalogue. (mc-9833)
- Replace Subscribed Model `save` method with `federate`. Add parameters (which are not saved) to optionally specify which URL, content type, and/or ImporterProviderService to use when federating the model. Check to ensure the specified URL is published in the Subscribed Catalogue. (mc-9834)
- Combine `getFileType` and `getProducesContentType` methods in ExporterProviderService to `getContentType`. (mc-9835)
- Split SubscribedCatalogueFunctionalSpec tests into base tests and AtomSubscribedCatalogueFunctionalSpec and MauroJsonSubscribedCatalogueFunctionalSpec tests. Also add Atom and Mauro JSON tests for relevant SubscribedModelFunctionalSpec tests.


joe-crawford:
Failing test on branch and PR builds is not related.

joe-crawford:
Compatibility note:

- The changes are backwards incompatible for subscription to and from older versions of Mauro as subscribed catalogues (i.e. the new 'Mauro JSON' subscribed catalogue type is not compatible with the previous subscribed catalogues). The migration script does migrate previous subscribed catalogues if both remote and local instances are updated.
- The changes to the content type fields in ExporterProviderService are backwards incompatible.

Notes on changes:
- In SubscribedModelService, `federateSubscribedModel` supports federating an MdmDomain if certain properties are found dynamically. The intention is to allow federating a Model or a VersionedFolder, in particular so importers for FHIR JSON multi-item types (FHIR packages/bundles) can import a VersionedFolder and a 1:1 link between SubscribedModels and imported domains is maintained.
-  SubscribedModelFunctionalSpec tests check that this PR is present on the CD instance before running (by checking for SC `/types` endpoint).
- Re: mc-9835 - grails.mime.types definitions not used as these affect client-side content type negotiation. URLs are imported as byte arrays and passed to importers.

joe-crawford:
Failing PR test unrelated again.

mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/traits/provider/importer/XmlImportMapping.groovy
```diff
@@ -25,12 +25,12 @@ import groovy.xml.slurpersupport.NodeChildren
  */
 trait XmlImportMapping {
 
-    Map<String, Object> convertToMap(NodeChild nodes) {
+    Map<String, Object> convertToMap(NodeChild nodes, List<String> exclude = ['id']) {
```
olliefreeman, line 28: where are we excluding anything other than id??

mdm-plugin-federation/grails-app/utils/uk/ac/ox/softeng/maurodatamapper/federation/converter/SubscribedCatalogueConverter.groovy
```diff
@@ -0,0 +1,40 @@
...
+ */
+package uk.ac.ox.softeng.maurodatamapper.federation.converter
+
+import uk.ac.ox.softeng.maurodatamapper.core.authority.Authority
+import uk.ac.ox.softeng.maurodatamapper.federation.PublishedModel
+import uk.ac.ox.softeng.maurodatamapper.federation.SubscribedCatalogue
+import uk.ac.ox.softeng.maurodatamapper.federation.SubscribedCatalogueType
+import uk.ac.ox.softeng.maurodatamapper.federation.web.FederationClient
+
+interface SubscribedCatalogueConverter {
```
olliefreeman, line 26: can we use traits not j17 interfaces please

mdm-plugin-federation/grails-app/services/uk/ac/ox/softeng/maurodatamapper/federation/SubscribedModelService.groovy
```diff
@@ -47,10 +52,11 @@ import java.time.OffsetDateTime
 class SubscribedModelService implements SecurableResourceService<SubscribedModel>, AnonymisableService {
 
     @Autowired(required = false)
-    List<ModelService> modelServices
+    List<MdmDomainService> domainServices
```
olliefreeman, line 55: i would keep this at modelServices, as we're not going to federate and subscribe to anything other than models, which means youre just making more processing calculations as you have to search more objects to find a match

mdm-plugin-federation/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/federation/SubscribedModelController.groovy
```diff
@@ -50,17 +53,23 @@ class SubscribedModelController extends EditLoggingController<SubscribedModel> {
     }
 
     @Transactional
-    @Override
-    def save() {
+    def federate(SubscribedModelFederationParams subscribedModelFederationParams) {
```
olliefreeman, line 56: if you're going to use command objects which are validatable then you should call validate on the object before you do anything else and then if the object is invalid you can respond with the objects errors field. this will ensure thate the SM instance is there and that everythign you want/need is provided
```groovy
if (subscribedModelFederationParams.hasErrors() || !subscribedModelFederationParams.validate()) {
transactionStatus.setRollbackOnly()
respond subscribedModelFederationParams.errors // STATUS CODE 422
return 
}
```

mdm-plugin-federation/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/federation/SubscribedModel.groovy
```diff
@@ -47,7 +47,7 @@ class SubscribedModel implements MdmDomain, SecurableResource, EditHistoryAware
         subscribedCatalogue nullable: false
         folderId nullable: false
         subscribedModelId nullable: false, unique: 'subscribedCatalogue' // Should prevent subscribing to the same modelId from same catalogue
-        subscribedModelType blank: false
+        subscribedModelType nullable: true
```
olliefreeman, line 50: even with nullable:true you should leave blank:false in .. blank is a check against an empty string which we never want

mdm-plugin-federation/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/federation/SubscribedModel.groovy
```diff
@@ -30,7 +30,7 @@ class SubscribedModel implements MdmDomain, SecurableResource, EditHistoryAware
     UUID id
 
     //The ID of the model on the remote (subscribed) catalogue
-    UUID subscribedModelId
+    String subscribedModelId
```
olliefreeman, line 33: Why are we not storing this as a UUID??

olliefreeman (changes requested):
I would suggest adding a `public static final CONTENT_TYPE = ''` to each exporter service, and then having the getContentType() method returning this variable, then in the importer services you have the handlesContentType check against that variable, that way if we want to change the content type we only have to do it in 1 place and it applies across all things that use it. It also means if we have a importer which handles multiple types they can check a list of psf variables 

See the other comments and questions

mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/traits/provider/importer/XmlImportMapping.groovy
joe-crawford: This was part of an attempt to use convertToMap with Atom XML, where I wanted to include `id` elems, however not used because of data in XML attributes. I can revert this.

mdm-plugin-federation/grails-app/utils/uk/ac/ox/softeng/maurodatamapper/federation/converter/SubscribedCatalogueConverter.groovy
joe-crawford: Will change.

mdm-plugin-federation/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/federation/SubscribedModelController.groovy
joe-crawford: Will change.

mdm-plugin-federation/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/federation/SubscribedModel.groovy
joe-crawford: Will change.

mdm-plugin-federation/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/federation/SubscribedModel.groovy
joe-crawford: This is because the `subscribedModelId` is now taken from the ID field from a Mauro JSON catalogue (UUID) or the `id` from an Atom feed, which might not be a UUID.

mdm-plugin-federation/grails-app/services/uk/ac/ox/softeng/maurodatamapper/federation/SubscribedCatalogueService.groovy
```diff
@@ -79,27 +75,28 @@ class SubscribedCatalogueService implements XmlImportMapping, AnonymisableServic
 
     void verifyConnectionToSubscribedCatalogue(SubscribedCatalogue subscribedCatalogue) {
         try {
-            Map<String, Object> catalogueModels = getFederationClientForSubscribedCatalogue(subscribedCatalogue).withCloseable {client ->
-                client.getSubscribedCatalogueModels(subscribedCatalogue.apiKey)
-            }
-            if (!catalogueModels.containsKey('publishedModels') || !catalogueModels.authority) {
+            def (Authority subscribedAuthority, List<PublishedModel> publishedModels) = listPublishedModelsWithAuthority(subscribedCatalogue)
+
+            if (!subscribedAuthority.label || publishedModels == null) {
```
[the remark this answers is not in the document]
joe-crawford: Are we sure we want to reject the empty list as this will prevent subscribing to valid catalogues which haven't yet published models?

mdm-plugin-federation/grails-app/services/uk/ac/ox/softeng/maurodatamapper/federation/SubscribedModelService.groovy
olliefreeman: okay ... in that case leave as MdmDomainService list

mdm-plugin-federation/grails-app/services/uk/ac/ox/softeng/maurodatamapper/federation/SubscribedCatalogueService.groovy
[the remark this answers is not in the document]
olliefreeman: okay ... then use == null, however i would advise making a documented note/comment in the code that if that method returns null then it is invalid but it is allowed to return an empty list and be valid...with that justification...that way people will know why we're not using the standard groovy check

mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/traits/provider/importer/XmlImportMapping.groovy
joe-crawford: Reverted

mdm-plugin-federation/grails-app/utils/uk/ac/ox/softeng/maurodatamapper/federation/converter/SubscribedCatalogueConverter.groovy
joe-crawford: Changed to trait

mdm-plugin-federation/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/federation/SubscribedModelController.groovy
joe-crawford: Added the validation snippet, thanks.

mdm-plugin-federation/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/federation/SubscribedModel.groovy
joe-crawford: Changed

mdm-plugin-federation/grails-app/services/uk/ac/ox/softeng/maurodatamapper/federation/SubscribedCatalogueService.groovy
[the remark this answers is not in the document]
joe-crawford: Added comment

mdm-plugin-federation/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/federation/SubscribedModelController.groovy
```diff
@@ -50,17 +53,32 @@ class SubscribedModelController extends EditLoggingController<SubscribedModel> {
...
 
-        def instance = createResource()
+        if (subscribedModelFederationParams.hasErrors() || !subscribedModelFederationParams.validate()) {
+            transactionStatus.setRollbackOnly()
+            respond subscribedModelFederationParams.errors // STATUS CODE 422
+            return
+        }
+
+        // Validate nested command object separately
+        if (subscribedModelFederationParams.importerProviderService &&
```
olliefreeman, line 66: shouldnt need to do this, the validation should cascade?

olliefreeman (commented):
just the one question about the validation

mdm-plugin-federation/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/federation/SubscribedModelController.groovy
joe-crawford: It actually doesn't seem to cascade to the nested command object. It does to the domain object as expected though. I did try adding a custom validator in the SubscribedModelFederationParams to call the nested validate method which works but would require more handling to push the errors back up than the separate call.

Found this which seems to imply it's expected: https://plugins.grails.org/plugin/ctoestreich/cascade-validation (but surprising).

Issue #278, opened 2022-02-24
Cannot import data elements from the same VF


jamesrwelch:
When trying to 'import' a data element from the same, draft, versioned folder, an error is returned.  You should be able to import data elements from the same Model / Versioned Folder, or from another, finalised Model / VF.
```json
{
    "total": 1,
    "errors": [
        {
            "message": "DataElement [e2ae6e37-38ff-464d-bae9-b892632e0f0c] to be imported does not belong to a finalised DataModel"
        }
    ]
}
```

olliefreeman:
So the validation should check/allow for importing from the same DM even when the DM is not finalised. The same allowance needs to be extended for DMs which exist inside a VF.

olliefreeman:
this actually covers all types of imports, DCs and DTs into a DM and DCs and DEs into a DC

jamesrwelch:
Re-opened because this fails when the models are in different folders within the same VF.  Fix shortly in a PR

PR #328, opened 2022-06-24
Import Data Elements into Data Classes within the same VF

jamesrwelch:
Check against the top-level path item, rather than the immediate parent.  Solves #278 where the models are in different folders inside the VF


joe-crawford:
Todo: Add test for this branch

PR #332, opened 2022-07-09
gh-274 set referenceDataModel from URL

### LINKED ISSUE #274: Cannot create a new Ref Data Value using standard expected body
The following endpoint `{{base_url}}/referenceDataModels/{{reference_data_model_id}}/referenceDataValues` should accept the following body

```json
{
    "rowNumber": {{$randomInt}},
    "value": "{{$randomLoremSlug}}",
    "referenceDataElement": {
        "id": "{{reference_data_element_id}}"
    }
}
```

however its throwing a validation error

```json
{
    "total": 1,
    "errors": [
        {
            "message": "Property [referenceDataModel] cannot be null"
        }
    ]
}
```

which suggest the RDM_id is not being assigned from the URL when the object is created

aaronforshaw:
Closes #274

- Update functional tests, removing `referenceDataModel` from the body of each test, which has the effect of creating a failing test for the scenario described in the ticket
- Override the `createResource`method so that we can set the referenceDataModel using the ID from the URL, rather than wrongly expecting it to be in the body of the request


joe-crawford:
Noticed issue #334 while looking at this but that issue already existed on develop so will merge, thank you.

PR #333, opened 2022-07-11
Batch save multiple terminologies

jamesrwelch:
This PR adds the functionality to batch-save multiple terminologies at the same time.  The existing batch-save works on a single terminology, which brings a big speed improvement for a single terminology with many terms, but this new service method brings the same speed improvement when saving many small terminologies.


joe-crawford:
Test failures not related.

PR #336, opened 2022-07-11
Add a test for #328

jamesrwelch:
This change introduces a test for the code change of #328 (There were no tests originally for this functionality).
Adds to the fix for #278


joe-crawford:
Test failures don't seem related.

PR #337, opened 2022-07-13
gh-256 use the securableResourceId

### LINKED ISSUE #256: The list all user groups for SR by GR endpoint doesnt work
The following endpoint `/api/${securableResourceDomainType}/${securableResourceId}/groupRoles/${groupRoleId}/userGroups` is not assigning the securableResourceId correctly resulting in a null id being searched for

```json
{
    "path": "/api/folders/5bff4fef-42e2-49ca-8f0c-d957ccbf571c/groupRoles/c92b1f22-70d7-4b70-8862-32e4223b0f26/userGroups",
    "resource": "Folder",
    "id": "null"
}
```

aaronforshaw:
Closes #256 

Fall back to using `securableResourceId` and add a test for the endpoint `GET /{securableResourceDomainType}/{securableResourceId}/groupRoles/{groupRoleId}/userGroups`

### DIFF
--- a/mdm-security/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/security/UserGroupInterceptor.groovy
+++ b/mdm-security/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/security/UserGroupInterceptor.groovy
@@ -40,7 +40,12 @@ class UserGroupInterceptor extends TieredAccessSecurableResourceInterceptor {
 
     @Override
     UUID getId() {
-        params.userGroupId ?: params.id ?: params.securableResourceId
+        params.userGroupId ?: params.id
+    }
+
+    @Override
+    UUID getSecuredId() {
+        getId() ?: params.securableResourceId
     }
 
     boolean before() {
--- a/mdm-security/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/security/UserGroupInterceptor.groovy
+++ b/mdm-security/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/security/UserGroupInterceptor.groovy
@@ -40,7 +40,7 @@ class UserGroupInterceptor extends TieredAccessSecurableResourceInterceptor {
 
     @Override
     UUID getId() {
-        params.userGroupId ?: params.id
+        params.userGroupId ?: params.id ?: params.securableResourceId
     }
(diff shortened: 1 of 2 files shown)


joe-crawford:
Thanks for this @aaronforshaw, I've got one comment.

mdm-security/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/security/UserGroupInterceptor.groovy
```diff
@@ -40,7 +40,7 @@ class UserGroupInterceptor extends TieredAccessSecurableResourceInterceptor {
 
     @Override
     UUID getId() {
-        params.userGroupId ?: params.id
+        params.userGroupId ?: params.id ?: params.securableResourceId
     }
```
joe-crawford, line 44: I think this could be a bit confusing as normally the `getId` method should return the ID of the type corresponding to the Controller and `getSecuredClass` method, so UserGroup in this case.

Would you be able to change this to the below? The tests should pass with:

```groovy
@Override
UUID getId() {
    params.userGroupId ?: params.id //?: params.securableResourceId
}

boolean before() {
    checkIds() // instead of securableResourceChecks()
    if (isIndex()) {
```

joe-crawford (changes requested):
Thanks @aaronforshaw, I've added one comment.

mdm-security/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/security/UserGroupInterceptor.groovy
aaronforshaw: This didn't quite work as some other tests failed.

But I realised I could revert the change to `UUID getUUID()`and instead override `UUID getSecuredId()`.

I think the test failures on Jenkins are down to Jenkins, as I have run them locally and they pass.

joe-crawford: Thanks @aaronforshaw, good to clarify to reduce the chance of accidentally checking the wrong object's permissions if adding future endpoints.

joe-crawford (approved):

PR #346, opened 2022-07-21
Feature/gh 254 Fix creation of SecurableResourceGroupRole on UserGroup endpoint

### LINKED ISSUE #254: SRGR Create by UserGroup endpoint doesnt work
The endpoint ` /api/userGroups/${userGroupId}/securableResourceGroupRoles` does not work as intended, currently throws an exception

```json
{
    "status": 400,
    "reason": "Bad Request",
    "errorCode": "SRGRS01",
    "message": "SecurableResourceGroupRole retrieval for securable resource [UserGroup] with no supporting service",
    "path": "/api/userGroups/07cc0e4f-c356-4e9a-ae32-509fbcd1915f/securableResourceGroupRoles",
    "environment": "DEVELOPMENT",
    "version": "5.1.0-SNAPSHOT"
}
```

aaronforshaw:
Closes #254 

- Make a small fix to the SecurableResourceGroupRoleController and SecurableResourceGroupRoleService to allow the findSecurableResource method to return a null value rather than throw an exception. This fixes the original bug.
- As required by the ticket, also add E2E tests for SecurableResourceGroupRole. As I am not certain what the requirements are, the test expectations are set up to pass given the current implementation in code.
- One problem I did notice is that there seems to be no `/edits`endpoint for SecurableResourceGroupRole. To ensure that the test suite passes I've added a test expectation `hasEditsEndpoint` and used this to skip testing of edits for SecurableResourceGroupRole. If it is required to add an endpoint for retrieving edits then this could be raised as a separate ticket, and the change to test expectations could then be reverted.

### DIFF
--- a/mdm-security/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/security/role/SecurableResourceGroupRoleController.groovy
+++ b/mdm-security/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/security/role/SecurableResourceGroupRoleController.groovy
@@ -17,8 +17,8 @@
  */
 package uk.ac.ox.softeng.maurodatamapper.security.role
 
-
 import uk.ac.ox.softeng.maurodatamapper.core.controller.EditLoggingController
+import uk.ac.ox.softeng.maurodatamapper.security.SecurableResource
 import uk.ac.ox.softeng.maurodatamapper.security.UserGroupService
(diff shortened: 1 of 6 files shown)


joe-crawford (approved):
Great, thanks @aaronforshaw.

I think we're fine to leave edits off the SRGRs unless a use case for them comes up.

Tests passing locally.

Issue #357, opened 2022-08-16
Add API property to control default state of "includeModelSuperseded" filter


pjmonks:
Extracted from a comment in issue #288:

"At the same time we want to add an ApiProperty that is bootstrapped in mdm-core/mdm-core to value `false` which allows admin users to set the default for "includeModelSuperseded" which should be loaded inside TreeItemController and that value passed to the service if the query param is not supplied"

_Originally posted by @olliefreeman in https://github.com/MauroDataMapper/mdm-core/issues/288#issuecomment-1065225054_

PR #358, opened 2022-08-17
gh-288 Update tree endpoints to filter superseded Versioned Folders

### LINKED ISSUE #288: Allow filtering superseded versioned folders in tree
# Description

The tree endpoints include parameters for filtering models to only latest versions (`includeModelSuperseded` and `includeDocumentSuperseded`). Versioned Folders should also be allowed to be filtered in the tree using these parameters.

# Additional context

From discussion: `includeModelSuperseded` should be used for both models and versioned folders. 

Related to: maurodatamapper/mdm-ui#490, adding these params would allow the UI option to apply to versioned folders as well.

pjmonks:
Resolves #288 

Update the tree endpoints, like `GET api/tree/folders/{id}` to use the provided `includeModelSuperseded` and `includeDeleted` parameters on Versioned Folders. By default Versioned Folders that are superseded are _not_ included in the tree response, unless `includeModelSuperseded=true` is provided.

Tests around the `VersionedFolderService` and `TreeItemService` have also been updated to test the behaviour.

This is my first PR to `mdm-core`, and also my first experience working in Groovy, so any feedback would be welcome.

### DIFF
--- a/mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/container/VersionedFolderService.groovy
+++ b/mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/container/VersionedFolderService.groovy
@@ -1197,8 +1197,8 @@ class VersionedFolderService extends ContainerService<VersionedFolder> implement
         savedCopy
     }
 
-    List<VersionedFolder> filterAllReadableModels(Collection<VersionedFolder> containers, boolean includeDocumentSuperseded,
-                                                  boolean includeModelSuperseded, boolean includeDeleted) {
+    List<VersionedFolder> filterAllReadableContainers(Collection<VersionedFolder> containers, boolean includeDocumentSuperseded,
+                                                      boolean includeModelSuperseded, boolean includeDeleted) {
         List<UUID> ids = containers.findAll {includeDeleted ? true : !it.deleted}.collect {it.id}
         List<UUID> constrainedIds
         // The list of ids are ALL the readable ids by the user, no matter the model status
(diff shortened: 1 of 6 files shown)


joe-crawford:
@pjmonks, sorry for the delay. I've tested this and it works great.

My only comment is that there are is some odd uneven reformatting, e.g. brackets not on matching lines etc. Maybe the codestyle XML file was not enabled? Due to project status I'll update the formatting now.

joe-crawford:
I agree with splitting out #357 as that can come separately. As there is the user preferences level setting for the superseded models I'd guess it's a low priority. I did notice that checking/unchecking the user preferences setting in the UI then going to look at the tree view didn't always have the expected effect but I think that's probably in the UI.

mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/container/VersionedFolderService.groovy
```diff
@@ -275,19 +274,18 @@ class VersionedFolderService extends ContainerService<VersionedFolder> implement
             long st = System.currentTimeMillis()
             Collection<Model> modelsInFolder = service.findAllByFolderIdInList(foldersInside)
             log.debug('Found {} {} inside VF', modelsInFolder.size(), service.getDomainClass().simpleName)
-            modelsInFolder.each {model ->
-                service.finaliseModel(model as Model, user, folderVersion, null, folderVersionTag)
+            modelsInFolder.each {model -> service.finaliseModel(model as Model, user, folderVersion, null, folderVersionTag)
             }
```
joe-crawford, line 278: Most of the reformatting is fine, just lines like this I'll reformat.

joe-crawford (approved):

mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/container/VersionedFolderService.groovy
```diff
@@ -1217,4 +1196,23 @@ class VersionedFolderService extends ContainerService<VersionedFolder> implement
...
         savedCopy
     }
+
+    List<VersionedFolder> filterAllReadableContainers(Collection<VersionedFolder> containers, boolean includeDocumentSuperseded,
+                                                      boolean includeModelSuperseded, boolean includeDeleted) {
+        List<UUID> ids = containers.findAll {includeDeleted ? true : !it.deleted}.collect {it.id}
+        List<UUID> constrainedIds
+        // The list of ids are ALL the readable ids by the user, no matter the model status
+        if (includeDocumentSuperseded && includeModelSuperseded) {
+            constrainedIds = new ArrayList<>(ids)
```
joe-crawford, line 1206: Usually don't need to specify ArrayList, LinkedHashMap etc. explicitly in Groovy.

Issue #363, opened 2022-09-14
API endpoint to get the latest finalised data model by data model label


GBishop-PHSA:
# Description

The API currently provides a Get request to return the latest data model using the data model path (label and optional version number). When the version number is excluded from the path, it returns draft models, if one exists for that label. An additional request parameter to limit results to the latest finalised data model would be helpful for sharing links to published data models through email or internal documents.

# Additional context

Current API request to find latest data model by label: 
http://localhost:4200/#/catalogue/dataModel/dm:PATH
* where PATH can also include a specific version number,  dm:SampleModel$1.0.0

New request parameter for API request to find latest finalised data model by label:
http://localhost:4200/#/catalogue/dataModel/dm:PATH?finalised=true
* where PATH does not include a version number,  dm:SampleModel?finalised=true

GBishop-PHSA:
Ready for review
mdm-ui PR: https://github.com/MauroDataMapper/mdm-ui/pull/665
mdm-core PR: https://github.com/MauroDataMapper/mdm-core/pull/364

joe-crawford:
Closing as completed.

PR #364, opened 2022-09-28
gh-363 request param to select finalized models by label

GBishop-PHSA:
gh-363 Provide a path in the API to view the latest finalised datamodel

eg. http://localhost:8080/api/dataModels/path/dm:DM_LABEL?finalised=true

### DIFF
--- a/mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/path/PathController.groovy
+++ b/mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/path/PathController.groovy
@@ -56,9 +56,10 @@ class PathController extends RestfulController<CatalogueItem> implements MdmCont
             }
 
             // Permissions have been checked as part of the interceptor
-            pathedResource = pathService.findResourceByPathFromRootResource(resource as MdmDomain, params.path)
+            pathedResource = pathService.findResourceByPathFromRootResource(resource as MdmDomain, params.path, null, [finalised: params.finalised])
         } else {
-            pathedResource = pathService.findResourceByPathFromRootClass(params.securableResourceClass, params.path, currentUserSecurityPolicyManager)
+            pathedResource =
+                pathService.findResourceByPathFromRootClass(params.securableResourceClass, params.path, currentUserSecurityPolicyManager, [finalised: params.finalised])
         }
 
         if (!pathedResource) return notFound(DomainClass, params.path)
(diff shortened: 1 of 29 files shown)


joe-crawford:
@GBishop-PHSA, thanks, this looks like a good solution, however there's some errors in the tests relating to the path endpoints for folders. I think this may need some more handling for Versioned Folders (which should work like DataModels).

https://jenkins.cs.ox.ac.uk/job/Mauro%20Data%20Mapper/job/mdm-core/job/PR-364/1/display/redirect

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/path/PathController.groovy
```diff
@@ -58,7 +58,8 @@ class PathController extends RestfulController<CatalogueItem> implements MdmCont
             // Permissions have been checked as part of the interceptor
             pathedResource = pathService.findResourceByPathFromRootResource(resource as MdmDomain, params.path)
```
joe-crawford, line 59: Does this method `pathService.findResourceByPathFromRootResource` also need changing as it looks like a similar route into looking up a domain by its path?

joe-crawford (changes requested):

GBishop-PHSA:
@joe-crawford I'm having an issue getting the tests to fail locally so I'm not sure how to investigate the failures

./gradlew -Dgradle.functionalTest=true :mdm-core:integrationTest
3 failed tests in AsyncJobFunctionalSpec

./gradlew -Dgradle.functionalTest=true :mdm-plugin-datamodel:integrationTest
1 failed test in DataModelFunctionalSpec

joe-crawford:
Thanks @GBishop-PHSA, this looks good and I'll merge it. I've made some minor tweaks in a branch `feature/gh-363` on this repo, could you pull those to your branch then I'll merge the PR?
- Auto reformat using the IntelliJ code style XML file
- Add extra abstract method without the params arg

The test failures originally were all HibernateOptimisticLockingFailureException or StaleStateException, basically these are sometimes expected failures of the Hibernate optimistic locking strategy that in actual use would need to be retried. They're related to a performance/reliability trade off of Hibernate, if they occurred more often in actual use it would be a problem but usually these happen just on the test server. They've passed after rerunning Jenkins.

GBishop-PHSA:
@joe-crawford, Thanks for looking at this. I've pulled in your changes and squashed the commits.

joe-crawford:
@GBishop-PHSA, Great, thank you!

joe-crawford (approved):

PR #368, opened 2022-11-03
Handle reordering of ModelItems depending on direction of reordering

### LINKED ISSUE #367: Sorting data elements results incorrect indexes
# Description

When reordering data elements, sometimes the resulting indexes are not unique which results in an incorrect order being displayed.

# Steps to reproduce

e.g.
1. Create a data class
2. Create data elements with labels `DE1`, `DE2`, `DE3`
3. Using UI (data class elements page) reorder DE1 to be in the second position
4. Refresh page and see the original order is still incorrectly displayed.

# Expected behavior

5. In database, the indexes seen are 1, 2, 2, for `DE1`, `DE2`, `DE3`, they should be 1, 0, 2. 

# Environment

Found while testing feature/gh-662 of mdm-core and mdm-ui but think the issue is in develop.

joe-crawford:
When reordering a ModelItem in a collection, take into account the direction in which the ModelItem is moved when reordering the siblings. Add a test for this. Resolves #367.


jamesrwelch (approved):
Not tested, but the code change makes sense, and if the test passes, then it's working better than it was before

PR #373, opened 2022-11-25
Check providerHasSavedModels parameter in async Model importer job

joe-crawford:
Don't revalidate model when imported from a provider which sets `providerHasSavedModels = true`.

When `providerHasSavedModels` is set the importer provider service should:

- Do all validation of the model
- Do all saving of the model
- Set any additional properties from the importer parameters, e.g. the folderId, itself.


jamesrwelch (approved):
This looks entirely sensible 👍

Issue #375, opened 2022-11-25
Allow results to be case-insensitive sorted


GBishop-PHSA:
# Description

Currently, if a GET request has a sort parameter set, the order is case sensitive. This can make it difficult to find an item, if data classe are always lower case but data models are always in upper case. A request should be able to specify if the results should have a case insensitive sort order.

# Additional context
The new feature could read a new request parameter called ignoreCase. If set to true, the sort will be case insensitive. If the ignoreCase parameter is excluded or set to false, the sort will be the default case sensitive order.

This is needed for the Classifier Elements screen, as the results include multiple catalogue item types which may have different naming standards (Sentence case or all lower case or all upper case).

GBishop-PHSA:
Pull requests:
https://github.com/MauroDataMapper/mdm-core/pull/376
https://github.com/MauroDataMapper/mdm-ui/pull/678

PR #376, opened 2022-11-26
gh-375 Add parameter to ignore case during sorting

GBishop-PHSA:
gh-375 Added a new paramter 'ignoreCase' so that sort order can be done case insensitive.

### DIFF
--- a/mdm-common/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/hibernate/search/PaginatedHibernateSearchResult.groovy
+++ b/mdm-common/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/hibernate/search/PaginatedHibernateSearchResult.groovy
@@ -61,6 +61,7 @@ class PaginatedHibernateSearchResult<K> {
         Integer offsetAmount = pagination.offset?.toInteger()
         String sortKey = pagination.sort ?: 'label'
         String order = pagination.order ?: 'asc'
+        Boolean ignoreCase = pagination.ignoreCase ? pagination.ignoreCase.toBoolean() : true
 
         List<D> sortedList = new ArrayList<>(fullResultSet)
         if ('distance' == sortKey) {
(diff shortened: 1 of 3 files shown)


joe-crawford:
Hi Gail, thanks for the contribution!

mdm-common/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/hibernate/search/PaginatedHibernateSearchResult.groovy
```diff
@@ -61,6 +61,7 @@ class PaginatedHibernateSearchResult<K> {
         Integer offsetAmount = pagination.offset?.toInteger()
         String sortKey = pagination.sort ?: 'label'
         String order = pagination.order ?: 'asc'
+        Boolean ignoreCase = pagination.ignoreCase ?: false
```
joe-crawford, line 64: Elsewhere we are actually sorting with a default of `ignoreCase=true`(in CatalogueItemService::applyHQLSort which affects various collections and also the Tree endpoints are case insensitive, at least for models), so I think this should default to true.

I think the check should be `Boolean ignoreCase = pagination.ignoreCase ? pagination.ignoreCase.toBoolean() : true`, the toBoolean is needed to convert the string `'false'` to Boolean false.

mdm-common/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/hibernate/search/PaginatedHibernateSearchResult.groovy
```diff
@@ -75,6 +76,14 @@ class PaginatedHibernateSearchResult<K> {
...
                     log.warn('Unknown sortfield for distance sorting {}', sortField)
             }
+        } else if (sortKey && ignoreCase) {
+            sortedList = fullResultSet.sort {a, b ->
+                if (order == 'asc') {
+                    a."$sortKey".compareToIgnoreCase(b."${sortKey}")
+                } else {
+                    b."$sortKey".compareToIgnoreCase(a."${sortKey}")
+                }
+            }
```
joe-crawford, line 86: I think this should be
```groovy
sortedList = fullResultSet.sort {a, b ->
if (order == 'asc') {
    a."$sortKey".compareToIgnoreCase(b."${sortKey}") ?: a."$sortKey".compareTo(b."${sortKey}")
} else {
    b."$sortKey".compareToIgnoreCase(a."${sortKey}") ?: b."$sortKey".compareTo(a."${sortKey}")
}
```
as otherwise the order isn't certain in the edge case where labels are equal except for different cases.

joe-crawford (changes requested):
I have a few comments below.

Looking at this made me realise our existing sorting code in CatalogueItemService also has exactly the same issues with reading the ignoreCase flag and the case insensitive search so I will raise a separate issue and submit a PR to fix that.

joe-crawford (changes requested):
Looks like tests S04 and S05 are failing because of the change to the sort order. Other tests like S03 are still passing but I think that's because `verifyJsonResponse` is not checking the order of the items in the expected list.

joe-crawford (approved):

joe-crawford:
Great, thanks.

PR #378, opened 2022-12-02
Make case insensitive ordering fall back on ordering by case

### LINKED ISSUE #377: Case insensitive ordering of collections should be a total ordering
# Description

The collections returned by endpoints using the CatalogueItemSearch::applyHQLOrder sorting method may have ambiguous order when items have labels that are equal except in case and the sort is case insensitive (which is default).

In some cases a request with `order=asc` is not equal to the reverse order of the same request with `order=desc`.
# Steps to reproduce

Create a DataModel and create DataClasses with identical labels except for case, e.g. `class1`, `Class1`, etc.

List the DataClasses with the `GET api/dataModels/$dataModelId/dataClasses` endpoint, sometimes the order is not the reverse of the same request with `?order=desc`.

# Expected behavior

Ordering should be `ORDER BY lower(label) <order>, label <order>`.

# Environment

mdm-core latest develop
# Additional context

Found while reviewing at PR #375.

joe-crawford:
Make case insensitive label ordering the equivalent of `ORDER BY lower(label), label` in CatalogueItemService and TreeItemService.

Resolves #377.


joe-crawford:
@jamesrwelch Yes sorry, I wasn't sure I had caught all of these cases yet so will have another look at this soon.

PR #386, opened 2023-01-12
Update validation rule for semantic links

### LINKED ISSUE #383: Unable to create a DataFlow from one model to itself
There's a validation rule that prohibits links from one model to itself.  However, the result of this is that you cannot create a DataFlow from a model to itself, which is a valid thing to do.  The rule needs to be changed so that it doesn't apply to DataFlow links, and perhaps any link other than a 'refines' / 'does not refine' link, where it really doesn't make sense.

jamesrwelch:
Fixes #383


joe-crawford (approved):
Looks good to me :+1:

PR #396, opened 2023-02-16
Add new ApiProperty to disable revealing exceptions in error messages

### LINKED ISSUE #381: Add configuration to allow hiding stacktraces in error responses
# Description

From Zulip:

>Invoke the API on (for example) : "api/authentication/login", POST method, with Content-type "application/json", and body with "user". The response is something like this:
```
{
"status": 500,
"reason": "Internal Server Error",
"errorCode": "UEX--",
"message":"Unable to determine the current character..."
...
}.
```
>This message is showing lots of internal details. Ideally it should return Status Code = 500 and a Json message with : `"error message" : "Internal Server Error"`.

# Additional context

Many applications support hiding the stacktraces for security reasons in public-facing deployments. We could add a new configuration parameter that controls whether messages and stacktraces are returned. Ideally this should be a new ApiProperty to allow setting from the admin Configuration page in the web interface.

Relevant gson template is at `mdm-core/grails-app/views/errors/_exception.gson`

joe-crawford:
Resolves #381.

If ApiProperty `security.hide.exception` is present and set to true, exception stack traces, internal error codes and error messages are hidden. If not set, exceptions are shown as normal. This setting doesn't affect Grails validation errors, for example "field is required" messages.

### DIFF
--- a/mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/admin/ApiPropertyService.groovy
+++ b/mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/admin/ApiPropertyService.groovy
@@ -51,11 +51,11 @@ class ApiPropertyService {
         apiProperty.delete(flush: true)
     }
 
-    ApiProperty findByKey(String key) {
+    static ApiProperty findByKey(String key) {
         ApiProperty.findByKey(key)
     }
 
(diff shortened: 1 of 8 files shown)


joe-crawford:
Test failures on branch aren't related.

pjmonks (approved):
These changes look acceptable, I've also tested with mdm-ui to confirm this works 👍

PR #397, opened 2023-02-23
Update Grails to 5.3.2 and other dependency updates

joe-crawford:
- Update to Grails 5.3.2 and Groovy 3.0.11
- Update copyright headers to include NHS England


jamesrwelch (approved):
I've run this up locally and it all seems to work nicely.  5.3.0 here we go!

PR #400, opened 2023-03-09
Add search type option to metadata search endpoint

### LINKED ISSUE #399: Update profile search filters to determine criteria type
# Description

Part of MauroDataMapper/mdm-ui#774 requires some adjustments to the existing `ProfileFilter` class in `mdm-plugin-profile`. Update this `SearchParamFilter` to return different types of criteria for wider support.

# Current implementation

To use the `ProfileFilter`, a search request must contain a JSON body similar to below:

```json
{
  "profileFields": [
    {
      "metadataNamespace": "...",
      "metadataPropertyName": "...",
      "filterTerm": "..."
    }
  ]
}
```

The `ProfileFilter` will identify that a `profileFields` array exists in the search params map and iterate over the array creating a `phrase` criteria for this metadata key.

This has some disadvantages:

1. Using `phrase` means that the metadata namespace, key and value must _exactly_ match an existing value in the metadata table.
2. These inputs are lengthy - namespaces are long strings, keys could be anything, and values are ultimately free text strings. Matching these exactly will prove to be cumbersome.

# Suggested implementation

Deciding the correct namespace, key and value to use will be part of MauroDataMapper/mdm-ui#774. For value search though, it is suggested that `ProfileFilter` be updated to accept a new parameter that can choose the search criteria type. For instance `phrase` is useful for an exact match on a known value (such as an enumeration type), but `keyword` is more useful for free text input.

The JSON body could potentially include a parameter like this:

```json
{
  "profileFields": [
    {
      "metadataNamespace": "...",
      "metadataPropertyName": "...",
      "filterTerm": "...",
      "type": "phrase"  // phrase | keyword (default: phrase)
    }
  ]
}
```

This will provide more control over the search over all metadata.

joe-crawford:
When searching on metadata values, allow specifying a criteria `type`. The default is `phrase`, the existing search type. The type `query` is similar to the default search on Catalogue Items. The `contains` type searches for terms that contain all of the search terms.

For an example item with metadata defined:
```json
{
    "namespace": "ns",
    "key": "key",
    "value": "ABC_DEF"
}
```

The following search (POSTed to `/api/catalogueItems/search`) will find the item:

```json
{
    "profileFields": [
        {
            "metadataNamespace": "ns",
            "metadataPropertyName": "key",
            "filterTerm": "de",
            "type": "contains"
        }
    ]
}
```

The following searches won't find the item:

```javascript
{
    "profileFields": [
        {
            "metadataNamespace": "ns",
            "metadataPropertyName": "key",
            "filterTerm": "DEF",
            "type": "phrase" // won't find item as `value` tokens not split on `_`
        }
    ]
}
```

```javascript
{
    "profileFields": [
        {
            "metadataNamespace": "ns",
            "metadataPropertyName": "key",
            "filterTerm": "DEF",
            "type": "query" // won't find item as `value` tokens not split on `_`
        }
    ]
}
```

Resolves #399.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -527,7 +527,7 @@ pipeline {
             jacoco classPattern: '**/build/classes', execPattern: '**/build/jacoco/*.exec', sourceInclusionPattern: '**/*.java,**/*.groovy',
                    sourcePattern: '**/src/main/groovy,**/grails-app/controllers,**/grails-app/domain,**/grails-app/services,**/grails-app/utils'
             archiveArtifacts allowEmptyArchive: true, artifacts: '**/*.log'
-            zulipNotification(topic: 'mdm-core')
+//             zulipNotification(topic: 'mdm-core')
         }
     }
 }
--- a/mdm-plugin-profile/grails-app/utils/uk/ac/ox/softeng/maurodatamapper/profile/rest/transport/search/searchparamfilter/ProfileFilter.groovy
+++ b/mdm-plugin-profile/grails-app/utils/uk/ac/ox/softeng/maurodatamapper/profile/rest/transport/search/searchparamfilter/ProfileFilter.groovy
@@ -17,6 +17,7 @@
  */
 package uk.ac.ox.softeng.maurodatamapper.profile.rest.transport.search.searchparamfilter
 
+import uk.ac.ox.softeng.maurodatamapper.api.exception.ApiBadRequestException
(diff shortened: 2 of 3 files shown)


joe-crawford:
Failing test doesn't look related.

pjmonks (approved):
I think this looks good 👍 When testing it myself, I think this new filter types will support what is needed for better metadata search.

Approved ✔️

PR #402, opened 2023-03-13
gh-401 Create endpoint to return single ProfileProviderService

### LINKED ISSUE #401: Provide an endpoint to return a single ProfileProviderService
# Description

There are two additions to `mdm-plugin-profile` that I think would be beneficial. This will help with metadata search.

# Get single Profile Provider Service

There is currently only an endpoint to return a list of `ProfileProviderService` objects:

```
GET /profiles/providers
```

However, there is no endpoint to get a single `ProfileProviderService`. This should be added:

```
GET /profiles/providers/{profileNamespace}/{profileName}/{profileVersion?}
```

Where:

1. `profileNamespace` and `profileName` are mandatory, `profileVersion` is optional
2. Should return a single instance of the requested `ProfileProviderService`

This would then align with other REST endpoints by providing both an `index` and `show` controller action for providers.

# List profile provider services by latest version

The `GET /profiles/providers` endpoint will always return every `ProfileProviderService`, including multiple version, since `latestVersionByMetadataNamespace` is hardcoded to be `false`. `latestVersionByMetadataNamespace` should be provided as an optional query parameter to allow the caller to decide this.

pjmonks:
- Also give profile providers index optional parameter to only return latest versions
- Add tests to verify new endpoint

Resolves #401

### DIFF
--- a/mdm-plugin-profile/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/profile/ProfileController.groovy
+++ b/mdm-plugin-profile/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/profile/ProfileController.groovy
@@ -49,7 +49,17 @@ class ProfileController implements ResourcelessMdmController, DataBinder {
 
 
     def profileProviders() {
-        respond profileProviderServices: profileService.getAllProfileProviderServices(false)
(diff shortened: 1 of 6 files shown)


joe-crawford:
Thanks @pjmonks, the code changes look good to me :+1:

Could you update the tracked_endpoints.txt file mentioned in one of the failing tests? I'll add a Postman entry for the URL as well.

joe-crawford (approved):

PR #403, opened 2023-03-15
gh-394: Save Theme images in the database

### LINKED ISSUE #394: Save Theme images in the database
# Description

The Mdm Explorer is being updated to allow admin users to change colours, fonts and logos of the Mdm Explorer. The management of the themes will me handled in the Mdm UI using Api Properties. We also need to give the admin user the ability to upload an image that can be used as a logo. This image needs to be saved in the backend. This task is to carry out the backend changes required to support this.

# Additional context

We already have a database table to hold image files called `user_image_file`. At the moment this is closely tied to users. This needs to be made more generic to handle images associated with other types of data. 
1. Rename the `user_image_file` table to be `image_file`. (Rename all references in the code to make sure this stays as is)
2. Change the column `user_id` so that it is nullable. 
A script in `mdm-core/grails-app/conf/db/migration/core/` will be required to do this.

In `mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file` create two new files called `ThemeImageFileController.groovy` and `ThemeImageFileInterceptor.groovy`. These should be based on the existing `UserImageFileController.groovy` and `UserImageFileInterceptor.groovy` files. 

`ThemeImageFileController.groovy`: This needs to keep all the same methods as were copied from `UserImageFileController`, but the contents of those methods need to use an api key to identify the image that is being shown or updated or saved. This should be done by calling a new service called `ThemeImageFileService`, that will handle the processing. `ThemeImageFileService `should be based on `mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/file/UserImageFileService.groovy` and live in the same folder. This needs to be able to handle any api property name, where the value of the api property is the Id of the image file that has been uploaded. For the logo, the new api property key will be called `theme-image-header-logo`.

`ThemeImageFileInterceptor.groovy`: This needs to only allow admin users. (Talk to Joe Crawford for more info)

NigelPalmer:
This is the backend work required to allow Theme Images to be uploaded. Resolves gh-394.


mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
```diff
@@ -0,0 +1,199 @@
...
+
+    ThemeImageFileController() {
+        super(ThemeImageFile)
+    }
+
+    @Override
+    def show() {
+        ThemeImageFile resource = null
+        if (params.apiPropertyId && themeImageFileService.apiPropertyHasImage(Utils.toUuid(params.apiPropertyId))) {
+            resource = themeImageFileService.findByApiPropertyId(params.apiPropertyId)
```
pjmonks: For consistency sake, I'm not sure why the line above uses `Utils.toUuid()` yet this line ignores the hard conversion and uses type inference. Maybe this is typical in this codebase, but I would have thought either infer both or cast both instead.

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    @Transactional
+    @Override
+    def update() {
+        // Allow the UI to use PUT to create the image
+        // This allows them to not have to have various additional computations to determine if they have an actual image
+        if (params.apiPropertyId && !apiPropertyService.findById(Utils.toUuid(params.apiPropertyId))) {
+            notFound(params.id)
+            return
```
pjmonks: Should this line say `return notFound(params.is)` instead? At the moment, it seems the `return` is `void` whereas other conditions return a proper response.

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
```diff
@@ -0,0 +1,199 @@
...
+        }
+        if (params.apiPropertyId && themeImageFileService.apiPropertyHasImage(Utils.toUuid(params.apiPropertyId)))
```
pjmonks: Does `ThemeImageFileService` need both an `apiPropertyHasImage()` and `findByApiPropertyId()` method? I believe `apiPropertyHasImage()` calls `findByApiPropertyId()` and checks for null, so it's shorthand. I'm not too familiar with the coding styles/conventions in `mdm-core`, so my opinion is only subjective.

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
```diff
@@ -0,0 +1,199 @@
...
+            saveApiProperty(params.apiPropertyId.toString())
+        }
+
+        deleteResult
+    }
+
+    @Override
+    protected ThemeImageFile queryForResource(Serializable id) {
+        ThemeImageFile themeImageFile
+        if (params.containsKey('apiPropertyId')) {
```
pjmonks, line 132: What is the difference between this line and `if (params.apiPropertyId)`? Another inconsistency I see, but don't have enough Groovy knowledge to know if there is an actual difference.

mdm-core/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/core/file/ImageFile.groovy
```diff
@@ -0,0 +1,68 @@
...
+
+import java.util.regex.Pattern
+
+/**
+ * @since 07/02/2020
+ */
+@Resource(readOnly = false, formats = ['json', 'xml'])
+class ImageFile implements CatalogueFile, MdmDomain {
+
+    private static final Pattern PRECUSOR = ~/^data:image\/[^;]*;base64,?/
```
pjmonks: Typo - seems like it should be `PRECURSOR`

mdm-core/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/core/file/ImageFile.groovy
```diff
@@ -0,0 +1,68 @@
...
+
+import grails.rest.Resource
+
+import java.util.regex.Pattern
+
+/**
+ * @since 07/02/2020
+ */
+@Resource(readOnly = false, formats = ['json', 'xml'])
+class ImageFile implements CatalogueFile, MdmDomain {
```
pjmonks: Looks like good use of inheritance for these image types

mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileService.groovy
```diff
@@ -0,0 +1,76 @@
...
+
+    ThemeImageFile get(Serializable id) {
+        ThemeImageFile.get(id)
+    }
+
+    List<ThemeImageFile> list(Map args) {
+        ThemeImageFile.list(args)
+    }
+
+    List<ThemeImageFile> listWithPagination(Map pagination = [:]) {
```
pjmonks: `listWithPagination()` and `list()` look identical to me, is there a reason for both to exist?

mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileService.groovy
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    ThemeImageFile findByApiPropertyId(UUID apiPropertyId) {
+        ApiProperty apiProperty = ApiProperty.findById(Utils.toUuid(apiPropertyId))
+        findByApiProperty(apiProperty);
+    }
+
+    ThemeImageFile findByApiProperty(ApiProperty apiProperty) {
+        if (!apiProperty) return null
+        if (!apiProperty.value.matches("[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}")) return null
```
pjmonks: This check to see if a string represents a UUID feels like a utility function to me, either to create one or check one already exists. Then stating `if (!isUuid(apiProperty.value))` will read a lot clearer.

mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/controller/EditLoggingController.groovy
```diff
@@ -157,9 +168,14 @@ abstract class EditLoggingController<T> extends RestfulController<T> implements
...
     protected T saveResource(T resource) {
-        log.trace('save resource')
-        resource.save flush: true, validate: false
-        if (resource.instanceOf(EditHistoryAware) && !params.boolean('noHistory')) resource.addCreatedEdit(getCurrentUser())
+        try {
+            log.trace('save resource')
+            resource.save flush: true, validate: false
+            if (resource.instanceOf(EditHistoryAware) && !params.boolean('noHistory')) resource.addCreatedEdit(getCurrentUser())
+        } catch (Exception e) {
+            String error = e.message;
```
pjmonks: I'm not sure what this `catch` block is trying to achieve, the `String error` has no usage after this line. Perhaps this is some testing code left over?

mdm-core/src/test/groovy/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileServiceSpec.groovy
```diff
@@ -0,0 +1,153 @@
...
+        when:
+        ThemeImageFile themeImageFile = service.createNewFile(Paths.get('src/test/resources/userimagefile_string_content.txt'), reader1)
+        checkAndSave(themeImageFile)
+
+        then:
+        themeImageFile.id != null
+    }
+
+    /*
+        ThemeImageFile findByApiPropertyId(UUID apiPropertyId)
```
pjmonks: Should these comments be here?

mdm-security/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/security/UrlMappings.groovy
```diff
@@ -50,6 +50,10 @@ class UrlMappings {
                 }
             }
 
+            group '/theme', {
+                '/image'(single: 'themeImageFile', excludes: DEFAULT_EXCLUDES)
```
pjmonks: I'm a bit confused why this mapping for theme image files exists in `mdm-security`, especially as I can't see any controllers/interceptors in this plugin for it (again, might be my lack of Grails knowledge though). What does this endpoint do?

mdm-testing-functional/build.gradle
```diff
@@ -109,6 +109,7 @@ grails {
         implementation project(':mdm-security')
         implementation project(':mdm-plugin-dataflow')
         implementation project(':mdm-plugin-profile')
+        runtimeOnly 'uk.ac.ox.softeng.maurodatamapper.plugins:mdm-plugin-explorer:1.0.0-SNAPSHOT'
```
pjmonks: This looks like it was included for testing, it should probably be removed before merging.

pjmonks (changes requested):
There are a lot of changes here, especially having to make deep changes to existing functionality, good effort.

I've given some cursory feedback but I think someone like @joe-crawford , who has more experience, should give a more detailed review.

mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileService.groovy
joe-crawford: I think you can do !Utils.toUuid(...) to check if it's a UUID.

mdm-core/src/test/groovy/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileServiceSpec.groovy
NigelPalmer: No. Removed.

mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/controller/EditLoggingController.groovy
NigelPalmer: Yes, this is debug code that should have been removed. Reverted the method to it's previous state.

mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileService.groovy
NigelPalmer: Updated to use Utils.toUuid.

mdm-core/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/core/file/ImageFile.groovy
NigelPalmer: Typo inherited from original implementation. Now fixed.

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
NigelPalmer: Fixed to use toUuid.

mdm-core/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/core/file/ImageFile.groovy
NigelPalmer: Thank you. :-)

mdm-security/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/security/UrlMappings.groovy
NigelPalmer: I've removed this.

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
NigelPalmer: `if (params.apiPropertyId)` will return false if the `apiPropertyId` key is present in the map but it has a value of `null` or if the key is not present at all. `params.containsKey('apiPropertyId')` will return true when the key is present in the map, but has a `null` value. In this case, when the `apiPropertyId` param is supplied, then we want to try and use it, even if it is null. We don't want to ignore it and try and use `id` instead. So it is correct to be used here. I have changed another `containsKey` reference though that is less relevant.

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
NigelPalmer: Fixed to return a proper response and the correct id!

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
NigelPalmer: Removed ThemeImageFileService and rejigged some logic that was using it to work without it.

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
joe-crawford: I think that ideally if the API property itself doesn't exist, that should return an appropriate error message, I think you should be able to use a call to notFound(ApiProperty, params.apiPropertyId) to return in that case.

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
```diff
@@ -0,0 +1,199 @@
...
+        saveResult
+    }
+
+    @Transactional
+    @Override
+    def save() {
+        // If the apiPropertyId refers to a non existent property, do not proceed
+        if (params.apiPropertyId && !apiPropertyService.findById(Utils.toUuid(params.apiPropertyId))) {
+            notFound(params.id)
+            return
```
joe-crawford: As above, I think this could be `return notFound(ApiProperty, params.apiPropertyId)`.

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
```diff
@@ -0,0 +1,199 @@
...
+
+    @Override
+    protected ThemeImageFile saveResource(ThemeImageFile resource) {
+        try {
+            resource.save flush: true, validate: false
+        } catch(Exception e) {
+            String errorMessage = e.message
+            throw e
+        }
+    }
```
joe-crawford: Is this possibly from debugging?

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
```diff
@@ -0,0 +1,199 @@
...
+
+    @Transactional
+    @Override
+    def delete() {
+        Object deleteResult
+
+        // If the apiPropertyId refers to a non existent property, do not proceed
+        if (params.apiPropertyId && !apiPropertyService.findById(Utils.toUuid(params.apiPropertyId))) {
+            notFound(params.id)
+            return
```
joe-crawford: Ditto about the notFound being on the apiPropertyId

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileInterceptor.groovy
```diff
@@ -0,0 +1,40 @@
...
+    ThemeImageFileInterceptor() {
+        order = Ordered.LOWEST_PRECEDENCE
+    }
+
+    boolean before() {
+        Utils.toUuid(params, 'apiPropertyId')
+        // Interception should be done by the user providing plugin
+        if (params.apiPropertyId) {
+            return true
+        }
```
joe-crawford: It looks like there is no security on the update of the image using the `apiPropertyId` because of this, which could lead to a free for all! ;)

I think this should check for admin rights for everything except the show action.

mdm-core/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/core/file/ImageFile.groovy
```diff
@@ -0,0 +1,68 @@
...
+    @Override
+    String getDomainType() {
+        ImageFile.simpleName
+    }
+
+    @Override
+    String getPathPrefix() {
+        'if'
+    }
+
```
joe-crawford: Could this possibly be an abstract class? Just to prevent any accidental attempts in the future to try and save a plain ImageFile to the database. Also I think I would remove the pathPrefix here as only the concrete objects should be pathable.

joe-crawford:
Thanks for this @NigelPalmer, looks good! As there's quite a few changes I'm still having a look but have an additional comment:

* When I tried to upload an SVG image through the API property endpoint, it stored it fine, but when I tried to retrieve it, the content-type header was wrong which meant the browser couldn't display it. I think it should be possible to allow SVGs by adding to `getContentType` in `CatalogueFile.groovy`.

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileInterceptor.groovy
joe-crawford: Actually to be consistent with other endpoints, everything under the `/admin` mapping should be admin-only, and the non-admin `show` mapping should set an `openAccess = true` param so it can be checked in the interceptor. An example is `ApiPropertyInterceptor.groovy`.

mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileService.groovy
NigelPalmer: One has a default, one doesn't. I spoke to Joe and he said it was safe to add the default to the list function, so I have done that and deleted liwtWithPagination.

mdm-testing-functional/build.gradle
NigelPalmer: Already removed to get tests to run successfully.

NigelPalmer:
svg files now displayed correctly. This also required changes to the mdm-ui to get this to work properly. Please get the latest commits from https://github.com/MauroDataMapper/mdm-ui/pull/794/commits to test this.

joe-crawford (changes requested):

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
joe-crawford: I can see this was updated, but I think the `ApiProperty` arg should be included to get the correct class name in the error message.

mdm-core/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/core/file/ImageFile.groovy
joe-crawford: This method can return `null` to disable pathing if it's required.

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
```diff
@@ -0,0 +1,204 @@
...
+        {
+            ThemeImageFile themeImageFile = themeImageFileService.findByApiPropertyId(Utils.toUuid(params.apiPropertyId))
+            if (themeImageFile) {
+                return super.update(themeImageFile)
+            }
+        }
+        if (params.id) return super.update()
+
+        Object saveResult = super.save()
+        if (saveResult && saveResult.Id && params.apiPropertyId) {
```
joe-crawford: Looks like a typo on the .Id property

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
```diff
@@ -0,0 +1,204 @@
...
+        // Ensure that if the user already has an image then we update rather than add another
+        if (params.apiPropertyId) {
+            ThemeImageFile themeImageFile = themeImageFileService.findByApiPropertyId(Utils.toUuid(params.apiPropertyId))
+            if (themeImageFile) {
+                return super.update(themeImageFile)
+            }
+        }
+
+        Object saveResult = super.save()
+        if (saveResult && saveResult.Id && params.apiPropertyId) {
```
joe-crawford: Ditto about .Id

mdm-security/src/test/groovy/uk/ac/ox/softeng/maurodatamapper/security/file/ThemeImageFileControllerSpec.groovy
```diff
@@ -0,0 +1,299 @@
...
+        params.apiPropertyId = apiProperty.id
+
+        controller.apiPropertyService = Mock(ApiPropertyService) {
+            findById(_) >> { UUID id ->
+                apiProperty
+            }
+
+            save(_,_) >> { ApiProperty apiPropertyInput, User updatedBy ->
+                apiProperty = apiPropertyInput
+                apiProperty.lastUpdatedBy = updatedBy
```
joe-crawford: This test seems to be passing locally but is failing on Jenkins, the issue is that `lastUpdatedBy` should be an email address rather than a user object, I think `.lastUpdatedBy = updatedBy.emailAddress` should fix it.

mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/controller/EditLoggingController.groovy
```diff
@@ -129,6 +138,8 @@ abstract class EditLoggingController<T> extends RestfulController<T> implements
         deleteResource instance
 
         deleteResponse instance
+
+        instance
     }
```
joe-crawford: @NigelPalmer, sorry, after our discussion last week I said I thought that this was fine to override the EditLoggingController, however looking at this with a fresh pair of eyes I have realised that there are some issues.

There are issues with the Grails framework if the response is set (which happens in the updateResponse/deleteResponse methods), and then a value is returned in the controller action method. I think the returned value triggers some additional response processing which fails as a response has already been sent. If I delete an image then hit the `DELETE` endpoint again, I get the following error response and stack trace:

```
{
    "status": 500,
    "reason": "Internal Server Error",
    "errorCode": "UEX--",
    "message": "JS-- - Unknown error",
    "path": "/error",
    "environment": "DEVELOPMENT",
    "version": "5.4.0-SNAPSHOT"
}
```

```
Mar 20, 2023 2:24:14 PM org.apache.catalina.core.StandardWrapperValve invoke
SEVERE: Servlet.service() for servlet [grailsDispatcherServlet] in context with path [] threw exception [Could not resolve view with name 'delete' in servlet with name 'grailsDispatcherServlet'] with root cause
javax.servlet.ServletException: Could not resolve view with name 'delete' in servlet with name 'grailsDispatcherServlet'
        at org.springframework.web.servlet.DispatcherServlet.render(DispatcherServlet.java:1384)
        at org.springframework.web.servlet.DispatcherServlet.processDispatchResult(DispatcherServlet.java:1149)
        at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:1088)
        at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:964)
        at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1006)
        at org.springframework.web.servlet.FrameworkServlet.doDelete(FrameworkServlet.java:931)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:702)
        at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:883)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:779)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:227)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.<============-> 99% EXECUTING [8m 37s]
        at org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:53)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:189)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:162)
        at org.grails.web.servlet.mvc.GrailsWebRequestFilter.doFilterInternal(GrailsWebRequestFilter.java:77)
        at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilte
```

I imagine it might be possible to trigger a similar error with the update change as well, it might take multiple calls to get the stacktrace to show however. Due to the return issue it could affect other classes that use the method.

I think the best solution here is actually just to override the update() and delete() methods in the ThemeImageFileController, then you can do the updates within that when needed. ApiPropertyController and SubscribedCatalogueController are examples where it's already overridden.

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
```diff
@@ -0,0 +1,204 @@
...
+    protected ThemeImageFile updateResource(ThemeImageFile resource) {
+        resource.save flush: true, validate: false
+    }
+
+    @Override
+    protected void deleteResource(ThemeImageFile resource) {
+        serviceDeleteResource(resource)
+    }
+
+    private void saveApiProperty(String apiPropertyId, String value = null) {
```
joe-crawford, line 179: Not very important but I think when this is called the values are originally always UUIDs?

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
```diff
@@ -0,0 +1,204 @@
+/*
+ * Copyright 2020-2023 University of Oxford and Health and Social Care Information Centre, also known as NHS Digital
```
joe-crawford: Some of the files need the licence changing to `NHS England` to get Jenkins to pass:

`[2023-03-20T10:57:39.816Z] Missing header in: mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
[2023-03-20T10:57:39.816Z] Missing header in: mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileInterceptor.groovy
[2023-03-20T10:57:39.816Z] Missing header in: mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileService.groovy
[2023-03-20T10:57:39.816Z] Missing header in: mdm-core/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/core/file/ImageFile.groovy
[2023-03-20T10:57:39.816Z] Missing header in: mdm-core/grails-app/domain/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFile.groovy
`

mdm-security/grails-app/views/themeImageFile/index.gson
```diff
@@ -0,0 +1,12 @@
...
+import grails.gorm.PagedResultList
+
+model {
+    Iterable<ThemeImageFile> themeImageFileList
+}
+
+json {
+    count themeImageFileList instanceof PagedResultList ? ((PagedResultList) themeImageFileList).getTotalCount() : themeImageFileList?.size() ?: 0
+    items tmpl.themeImageFile(themeImageFileList ?: [])
+}
```
joe-crawford, line 12: Are the views in mdm-security needed?

I'm also not sure that the index view is ever used however probably doesn't do any harm to leave it in mdm-core.

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
NigelPalmer: ApiProperty added

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
NigelPalmer: Again, ApiProperty added as suggested

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
NigelPalmer: Yes. Removed debug code.

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
NigelPalmer: Typo fixed.

mdm-security/src/test/groovy/uk/ac/ox/softeng/maurodatamapper/security/file/ThemeImageFileControllerSpec.groovy
NigelPalmer: Updated as suggested

mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/controller/EditLoggingController.groovy
NigelPalmer: I've removed the return value of instance from save, update and delete in EditLoggingController, and replicated the new versions of these methods at the higher level. I have also removed the default parameter.

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
NigelPalmer: No. When saveApiProperty is called from delete the value is not passed in because there is no value (it's been deleted). In that case the value is null and the apiProperty value will be saved as "USE DEFAULT IMAGE". This is important for the fornt end to know whether it is loading a default or not.

joe-crawford: Ah yes, just meant that the function type could use UUIDs as 'USE DEFAULT IMAGE' is inside here.

mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/controller/EditLoggingController.groovy
joe-crawford: Great, thanks, sorry about the confusion!

mdm-security/grails-app/views/themeImageFile/index.gson
NigelPalmer: Deleted security views. The index view was added because listAllReadableResources needed functionality added to it so it made sense to tie it all together, even though it is not currently used, it will not do any harm and may prove beneficial in the future.

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
NigelPalmer: Licenses updated

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
NigelPalmer: Typo fix

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
NigelPalmer: Updated as suggested

mdm-security/grails-app/views/themeImageFile/index.gson
joe-crawford: Great, makes sense :+1:

mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/controller/EditLoggingController.groovy
joe-crawford: Sorry @NigelPalmer I don't think the latest changes resolve this issue. I'm still getting the same error response and seeing the stacktrace after uploading a theme image file, then hitting the delete endpoint for it three times.

The issue is that the saveInstance/updateInstance/deleteInstance methods call the saveResponse/updateResponse/deleteResponse controller methods, which issue an HTTP response. If these issue a response, then the main controller method also tries to respond, the servlet gets into an inconsistent state. Also note the `validateResource` method also can issue a response and so can the `notFound` call so to make this safe, I think the only solution is make sure any validation is in the top level save/update/delete controller method, make sure there is always a return from the controller when any `notFound` is used, and don't call any response method inside the `*Instance` methods; these should just do the relevant action only.

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
```diff
@@ -201,4 +196,71 @@ class ThemeImageFileController extends EditLoggingController<ThemeImageFile> {
...
+        }
+
+        if (!validateResourceDeletion(instance, 'delete')) return
+
+        deleteResource instance
+
+        deleteResponse instance
+
+        instance
+    }
```
joe-crawford: Please see comment above about the `*Instance` methods calling the `*Response` methods.

joe-crawford (changes requested):

mdm-testing-functional/src/integration-test/groovy/uk/ac/ox/softeng/maurodatamapper/testing/functional/core/file/ThemeImageFileFunctionalSpec.groovy
```diff
@@ -0,0 +1,266 @@
...
+        }
+    }
+
+    @Transactional
+    def checkAndSetupData() {
+        log.debug('Check and setup test data')
+        sessionFactory.currentSession.flush()
+        getOrCreateApiProperties()
+        assert CatalogueUser.count() == 10
+        assert ApiProperty.count() == 45
```
joe-crawford: This is giving
```
Condition not satisfied:

ApiProperty.count() == 45
|           |       |
|           20      false
class uk.ac.ox.softeng.maurodatamapper.core.admin.ApiProperty

File: mdm-security/src/test/groovy/uk/ac/ox/softeng/maurodatamapper/security/test/SecurityUsers.groovy [hunk 27]
```diff
@@ -21,8 +21,8 @@ import uk.ac.ox.softeng.maurodatamapper.security.utils.SecurityDefinition
 
 trait SecurityUsers extends SecurityDefinition {
 
-    void implementSecurityUsers(String creatorKey) {
-        createModernSecurityUsers(creatorKey)
+    void implementSecurityUsers(String creatorKey, Boolean includeAdmin = false) {
+        createModernSecurityUsers(creatorKey, includeAdmin)
```
joe-crawford: This change is making the security unit tests fail, could you revert if not needed.

NigelPalmer: Apologies. I meant to revert that since it is not required. Now reverted.

mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/controller/EditLoggingController.groovy
NigelPalmer: I have reworked the code so that the methods always return a HTTP response.

mdm-security/grails-app/views/themeImageFile/index.gson
NigelPalmer: Re-added security views. The tests show that the code is insisting on using the security views. This is a bit odd. For the sake of getting this code complete, I think we should leave this in place for now and come back to it later. If you agree?

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/file/ThemeImageFileController.groovy
NigelPalmer: Reworked code to always return a HTTP Response.

mdm-testing-functional/src/integration-test/groovy/uk/ac/ox/softeng/maurodatamapper/testing/functional/core/file/ThemeImageFileFunctionalSpec.groovy
NigelPalmer: I was including a reference to mdm-plugin-explorer in my local build. This was causing the tests to run with different data. I removed the reference, and then was able to align the tests with how they run in Jenkins.

mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/controller/EditLoggingController.groovy
joe-crawford: @NigelPalmer, great, thanks for that.

mdm-security/grails-app/views/themeImageFile/index.gson
joe-crawford: @NigelPalmer, yes I agree, definitely fine to leave this for now :)

Perhaps all the tests in security for this could actually be in core but I don't think it's doing any harm and can see they are all passing :)

mdm-security/src/test/groovy/uk/ac/ox/softeng/maurodatamapper/security/test/SecurityUsers.groovy
joe-crawford: No worries.

joe-crawford (approved):
@NigelPalmer, thanks for this great piece of work which has taken a lot of effort! All tests are passing, and the code and API all looks good :+1:

PR #407, opened 2023-03-30
gh-406 Return path in each catalogue item GET endpoint

### LINKED ISSUE #406: Get path information from a catalogue item
# Description

There are endpoints that exist so far to return a catalogue item when provided with a path:

```
# Get a catalogue item by path. Define the root domain to search under e.g. `dataModels`
GET /{domainName}/path/{path}

# Get a catalogue item by path, searching directly under a known catalogue item (domain/id)
GET /{domainName}/{id}/path/{path}
```

However, there currently isn't an inverse to this - return a path from a catalogue item. This will be required to carry out the front end change MauroDataMapper/mdm-ui#427, since at the moment the Mauro Data Mapper UI is doing manual string manipulation to build these paths (and incorrectly).

# Additional context

All catalogue items store a `path` property in the database, so this could be returned in each index/show endpoint for each domain type.

For simplicity, the following endpoints/domains should be prioritised:

- Tree endpoints
- Data Models - including Data Classes, Data Elements and Data Types
- Terminologies - including Terms and Code Sets
- Reference Data Models - including Reference Data Types, Reference Data Elements and (possibly) Reference Data Values
- Containers - Folders, Versioned Folders and Classifiers

The list above should cover most needs in the UI to cross-reference other items in the catalogue for now.

**Note:** I've also noticed that calling this endpoint returns a `500 Internal Server Error`:

```
GET /classifiers/path/cl%3ADevelopment%20Classifier

SEVERE: Servlet.service() for servlet [grailsDispatcherServlet] in context with path [] threw exception [Request processing failed; nested exception is grails.views.ViewException: Error rendering view: Error rendering view: Template not found for name classifier] with root cause
grails.views.ViewException: Template not found for name classifier
```

pjmonks:
- Recorded path information for every catalogue item can now be used by callers
- Covers every major domain type plus tree endpoints
- Update tests for general catalogue item rendering
- Fix _classifier_full view to inherit base template correctly. This was not working when rendering from the PathController
- Classifier views now also return the `domainType` property to be consistent with all other catalogue items

Resolves #406 

The majority of these changes are actually test specifications to make sure all expected output is returned correctly. The actual changes are found in these views:

- _baseCatalogueItem.gson
- _classifier.gson
- _classifier_full.gson
- _folder.gson
- _treeItem.gson


joe-crawford (approved):
Thanks for this @pjmonks, a really big effort but can see that this will be useful :+1:

Issue #408, opened 2023-03-31
Folder Path Search: Finalised param being ignored


NigelPalmer:
# Description

The api end point http://localhost:8080/api/folders/path/{path}?finalised=true has a bug where it does not pass the finalised param on to other functions that require it. This can result in an unfinalised data model being returned.

# Steps to reproduce

e.g.
1. Open Mdm UI and login as an admin
2. Make sure that you have a finalised dataModel and an open dataModel .
3. Open a new tab in the same browser and go to http://localhost:8080/api/folders/path/{path}?finalised=true (e.g: [http://localhost:8080/api/folders/path/fo:OUH%20Data%20Warehouse%7Cdm:modules?finalised=true ](http://localhost:8080/api/folders/path/fo:OUH%20Data%20Warehouse%7Cdm:modules?finalised=true) )
4. Check the response. This will return the open dataModel instead of the finalised one: `...,"finalised":false,...`
5. See error

# Expected behavior

When calling http://localhost:8080/api/folders/path/{path}?finalised=true a finalised dataModel should be returned.

# Additional context

The fix is to simply pass the finalised param into the next function, then everything works as expected.

Update file: `mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/path/PathService.groovy`
Change line 149 to be: `findResourceByPathFromRootResource(rootResource, path, null, pathParams)`

PR #409, opened 2023-03-31
gh-408: Folder Path Search: Pass finalised param onto other functions

NigelPalmer:
Fulfils #408


joe-crawford (approved):
Thanks for this @NigelPalmer, looks like a sensible change to me!

Issue #411, opened 2023-04-17
Migration error when starting backend with a fresh database


abwilson23:
A table change in a migration in the core project is causing an error in a migration applied later which is contained in the security project. 
The error can be reproduced by:

1. Dropping the maurodatamapper database
2. Creating a new one
3. Running `gradle bootrun` in mdm-testing-functional. 

The message should be:

![image](https://user-images.githubusercontent.com/6440992/232539033-f16ed2b9-3c07-4cb4-9bd9-aab93676bca4.png)

joe-crawford:
Closed by PR 412.

PR #412, opened 2023-04-17
gh-411 - Fix migration error

abwilson23:
Update migration script fixes the issue when working with a fresh database. 
Also updated the security projects beforeValidate.sql script to ensure it works with existing databases.


joe-crawford (approved):
Thanks @abwilson23! This looks good to me :+1:

joe-crawford:
Can see failing tests are not related.

PR #421, opened 2023-06-08
Fix memory leak when importing multiple models

### LINKED ISSUE #420: Memory leak when importing new DataModels
# Description

Using a Python script with an API key, importing new models (one per request, using mdm-plugin-openclinica) resulted in an out of memory exception in the Tomcat process after a few thousand models were imported. After restarting the container, and carrying on importing models, the same happened after another few thousand imports. 

# Expected behavior

Importing models should not increase memory usage significantly.

# Environment

Please complete the following information:
- mdm-core latest develop
- mdm-plugin-openclinica latest develop

joe-crawford:
This PR prevents Grails/Hibernate from loading the `securableResourceGroupRoles` association on `UserGroup`s when `UserSecurityPolicyManager::addSecurityForSecurableResource` is called, which was causing a memory leak because the old `SecurableResourceGroupRole` objects were staying in memory after requests completed.

If code that calls `addSecurityForSecurableResource` later tries to check the same permissions on a UserGroup in the same request, the new security role might not appear there yet, but I think this is unlikely to be an issue.

I've tested in a local instance that the memory usage isn't increasing when importing multiple models with a script.

Resolves #420.

### DIFF
--- a/mdm-security/grails-app/services/uk/ac/ox/softeng/maurodatamapper/security/role/SecurableResourceGroupRoleService.groovy
+++ b/mdm-security/grails-app/services/uk/ac/ox/softeng/maurodatamapper/security/role/SecurableResourceGroupRoleService.groovy
@@ -82,7 +82,7 @@ class SecurableResourceGroupRoleService implements MdmDomainService<SecurableRes
             securableResource: securableResource,
             createdBy: createdBy.emailAddress
         )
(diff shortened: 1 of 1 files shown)


joe-crawford:
Failure on PR merge is a StaleStateException and I don't think it's related.

jamesrwelch (commented):
I'm happy with this - let's give it a go and see 👍

jamesrwelch (approved):

PR #423, opened 2023-06-15
Add SQL indexes to Data Model join tables

joe-crawford:
Improve performance by adding indexes on DataModel/facet join tables and breadcrumb tree columns. Speeds up some operations and the Lucene re-index.

- Indexes found by looking at long running queries and `pg_stat_user_tables` in an instance with many datamodels.
- Indexes on all the join table columns are needed because lookups happen on the datamodel/element/class ID and also when rows are deleted from the model/element/class table, a full table scan on the other columns (metadata ID etc.) can happen because of the foreign key constraints.

### DIFF
--- a/mdm-core/grails-app/conf/db/migration/core/V5_2_2__add_indexes.sql
+++ b/mdm-core/grails-app/conf/db/migration/core/V5_2_2__add_indexes.sql
@@ -1,0 +1,2 @@
+-- create index to speed up BreadcrumbTreeService::deleteAllByDomainIds
+CREATE INDEX breadcrumb_tree_tree_string_prefix_idx ON core.breadcrumb_tree (SUBSTR(tree_string, 1, 36+1));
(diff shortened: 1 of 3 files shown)


joe-crawford:
Failing test does not look related.

jamesrwelch (commented):
I've had a play - some definite speed-ups in a few places - thank you!

Do you think we need breadcrumb tree indexes on Terminology as well?  And how about on the model components?

joe-crawford:
I've updated to add a breadcrumb tree index for Terminologies, as it makes sense to be consistent with the other Models (CodeSet and DataModel) which had this added after profiling.

I haven't yet seen joins on the model item breadcrumb tree IDs during profiling, and these would be large indexes which would add to insertion time, so at the moment I'd suggest leaving these off unless we spot a problem there.

(Different failing test looks unrelated again.)

jamesrwelch (approved):
This is great - some definite speed-ups.  Sorry it's taken so long!

Issue #416, opened 2023-05-18
Add a filter in Search to hide superseded models


GBishop-PHSA:
# Description

Users of the Search page want to see search results of only the most recent versions of data models. A new filter option could be added to allow users to select whether they will see superseded versions or not, in the search results.

# Additional context

The search controller could be updated to accepts an additional parameter indicating whether superseded versions should appear in the search results. The UI would show this option on the Advanced Search as well as the Search Filter section of the search results page.

![SearchScreenShowSupercededFilter](https://github.com/MauroDataMapper/mdm-core/assets/90805521/8f793976-fa65-4e82-91df-15982527a8d4)
![SearchScreenShowSuperceded](https://github.com/MauroDataMapper/mdm-core/assets/90805521/451faaf5-7b1a-4bf3-9f46-6b8ac668cb2a)

GBishop-PHSA:
Pull requests for mdm-core and mdm-ui
https://github.com/MauroDataMapper/mdm-ui/pull/815
https://github.com/MauroDataMapper/mdm-core/pull/424

jamesrwelch:
@GBishop-PHSA Thank you so much for these PRs!  We'll review them in the next day for two :)

GBishop-PHSA:
@joe-crawford Thank you for making the changes to the pull requests. If they're both merged, can this ticket be closed?

joe-crawford:
Yes, sorry for the delay, closing as resolved by #424!

PR #424, opened 2023-06-15
GH-416 Add show superceded document search filter

GBishop-PHSA:
gh-416 Addition of a search parameter to determine whether superceded document version should be included in search results.

### DIFF
--- a/mdm-core/grails-app/utils/uk/ac/ox/softeng/maurodatamapper/core/rest/transport/search/SearchParams.groovy
+++ b/mdm-core/grails-app/utils/uk/ac/ox/softeng/maurodatamapper/core/rest/transport/search/SearchParams.groovy
@@ -35,7 +35,7 @@ class SearchParams implements Validateable {
     String sort
     String searchTerm
     Boolean labelOnly
-    Boolean supercededDocuments
+    Boolean includeSuperseded
     List<String> domainTypes
     Date lastUpdatedBefore
     Date lastUpdatedAfter
@@ -62,7 +62,7 @@ class SearchParams implements Validateable {
         classifiers = []
         classifierFilter = []
         labelOnly = false
-        supercededDocuments = false
+        includeSuperseded = false
     }
 
     void setLimit(Integer limit) {
--- a/mdm-core/grails-app/utils/uk/ac/ox/softeng/maurodatamapper/core/rest/transport/search/SearchParams.groovy
+++ b/mdm-core/grails-app/utils/uk/ac/ox/softeng/maurodatamapper/core/rest/transport/search/SearchParams.groovy
@@ -35,6 +35,7 @@ class SearchParams implements Validateable {
     String sort
     String searchTerm
     Boolean labelOnly
+    Boolean supercededDocuments
     List<String> domainTypes
     Date lastUpdatedBefore
     Date lastUpdatedAfter
@@ -61,6 +62,7 @@ class SearchParams implements Validateable {
         classifiers = []
         classifierFilter = []
         labelOnly = false
+        supercededDocuments = false
     }
 
     void setLimit(Integer limit) {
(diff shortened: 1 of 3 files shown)


mdm-core/grails-app/utils/uk/ac/ox/softeng/maurodatamapper/core/rest/transport/search/SearchParams.groovy
```diff
@@ -35,6 +35,7 @@ class SearchParams implements Validateable {
     String sort
     String searchTerm
     Boolean labelOnly
+    Boolean supercededDocuments
```
joe-crawford: Can I suggest this is changed to "includeSuperseded" rather than "supercededDocuments". Reason being is that in the past 'document version' was added as a separate version feature, now we are mainly trying to move away from using it and just use the main version functionality.

I can see this change looks at both the document version and the model version, which is good.

Minor point but we have `superseded` in the codebase already instead of `superceded`.

mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/search/AbstractCatalogueItemSearchService.groovy
```diff
@@ -93,17 +97,20 @@ abstract class AbstractCatalogueItemSearchService<K extends CatalogueItem> {
...
             }
+            results.removeIf {it.id in getExcludedCatalogueItemIds(searchParams, results.results)}
         } else {
             List<K> items = performSearch(owningIds, searchParams.searchTerm, searchParams.labelOnly,
                                           filteredDomainsToSearch, additional, customSearch)
             if (removeOwningIds) {
                 // Remove null entries and any which have an owning id, as we only want those inside the owners
                 items.removeIf {it.id in owningIds}
             }
+            items.removeIf {it.id in getExcludedCatalogueItemIds(searchParams, items)}
```
joe-crawford: I've tried this out on the default test/development DB:

When searching for "version tree", I get 9 models when 'Show superceded' = false, 13 when 'Show superceded' = true, which looks right.

When searching for "modify data class", I always get 13 data class results when 'Show superceded' is true and false. However some of these are contained within the superceded data models. I think the items in superceded parents should probably also be excluded? Probably something could be added to `getFilteredDomainsToSearch`.

mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/search/AbstractCatalogueItemSearchService.groovy
```diff
@@ -197,4 +204,20 @@ abstract class AbstractCatalogueItemSearchService<K extends CatalogueItem> {
         log.debug('Domain searching {}', domainToSearch)
         domainToSearch.customHibernateSearch(domainToSearch, owningIds, paths, pagination, additional, customSearch)
     }
+
+    protected List<K> getExcludedCatalogueItemIds(SearchParams searchParams, List<K> items) {
+        List<K> excludedIds
+        if (!searchParams.supercededDocuments) {
+            log.error("Superceded Param: {}", searchParams.supercededDocuments);
```
joe-crawford: Could you reduce the logging level to `log.debug`, thanks.

joe-crawford (changes requested):
Hi Gail, thanks a lot for this, definitely looks useful. Sorry for the delay, I've just got back from leave.

Could you have a look at the comments, mostly about the superceded owning items and update or let us know what you think.

GBishop-PHSA:
@joe-crawford  I've updated the search to remove catalogue items where the parent model is a superceded version and reduced the logging level

joe-crawford:
@GBishop-PHSA That's great. Thanks again for this contribution.

joe-crawford (approved):

Issue #415, opened 2023-05-18
Allow annotations to persist to a new version of a Data Model


GBishop-PHSA:
# Description
When creating a new draft data model from a previous version, the annotations (comments and attachments) are not copied to the new draft version.  

This feature could enabled/disabled through a new configuration property
Key: feature.copy_annotations_to_new_version    
Category: Features
Value: true

# Additional context
- the CatalogueItemService will check the config property and if true, copy all annotations(comments) and referenceFiles(attachements) to the new version

GBishop-PHSA:
Pull requests for this ticket
https://github.com/MauroDataMapper/mdm-ui/pull/816
https://github.com/MauroDataMapper/mdm-core/pull/425

joe-crawford:
Closing as resolved by #425.

PR #425, opened 2023-06-15
GH-415 Include comments and files during catalogue copy

GBishop-PHSA:
gh-415 Addition of copying files and comments when a new version of a catalogue item is created.

### DIFF
--- a/mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/model/CatalogueItemService.groovy
+++ b/mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/model/CatalogueItemService.groovy
@@ -17,6 +17,9 @@
  */
 package uk.ac.ox.softeng.maurodatamapper.core.model
 
+import uk.ac.ox.softeng.maurodatamapper.core.admin.ApiProperty
+import uk.ac.ox.softeng.maurodatamapper.core.admin.ApiPropertyService
+import uk.ac.ox.softeng.maurodatamapper.core.admin.ApiPropertyEnum
 import uk.ac.ox.softeng.maurodatamapper.core.container.Classifier
 import uk.ac.ox.softeng.maurodatamapper.core.container.ClassifierService
 import uk.ac.ox.softeng.maurodatamapper.core.diff.DiffCache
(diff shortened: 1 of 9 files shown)


mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/model/CatalogueItemService.groovy
```diff
@@ -176,6 +184,21 @@ abstract class CatalogueItemService<K extends CatalogueItem> implements MdmDomai
             semanticLinks = semanticLinkService.findAllBySourceMultiFacetAwareItemId(original.id)
         }
 
+        if (copyAnnotations()) {
+            annotations = annotationService.findAllByMultiFacetAwareItemId(original.id)
+            annotations.each {
+                copy.addToAnnotations(label: it.label, description: it.description,
+                        createdBy: copier.emailAddress, childAnnotations: it.childAnnotations)
```
joe-crawford: I don't think `childAnnotations: it.childAnnotations` will work because this will point to the old annotation IDs. I've tested this and it actually removes the child annotation from the original model and moves them to the copy. This would have to do a deep copy of the child annotations and create new child Annotation objects on the copy.

joe-crawford (changes requested):
Hi Gail, thanks for this PR, this looks really useful. Sorry again for the delay.

Could you have a look at the comment about the child annotations, thanks.

GBishop-PHSA:
@joe-crawford  I've updated the copy to call the annotation service deep copy method.

joe-crawford (approved):
Great, thanks for this feature :+1:

joe-crawford (changes requested):
Sorry @GBishop-PHSA just noticed there are some failing tests related to the Comments. Probably just needs the tests updating, could you have a look?

GBishop-PHSA:
@joe-crawford I'm not able to fix this test as it passes on my local computer:
CatalogueUserFunctionalSpec > Test exporting all Users to csv file (as admin) FAILED
    org.spockframework.runtime.ConditionNotSatisfiedError at CatalogueUserFunctionalSpec.groovy:1340

joe-crawford (approved):
Thanks for this. The tests are also passing for me locally.

I can't see the cause of the failing ones now but usually it's an unrelated Hibernate or concurrency issue.

PR #427, opened 2023-06-21
Delete facets belonging to terms when a terminology is deleted

### LINKED ISSUE #426: Term metadata is not removed when a terminology is deleted
When you delete a terminology, any metadata attached to terms in that terminology are not deleted.  This needs fixing, and any orphan metadata will need clearing in an update script

jamesrwelch:
Fixes #426


joe-crawford (approved):
Thanks @jamesrwelch, definitely looks good to me!

PR #431, opened 2023-07-24
gh-808 Added validation for max file size property

GBishop-PHSA:
gh-808 Is to add a helpful user message to the UI when a file that exceeds the maximum file size is set to be uploaded. 
Since this is set in the plugin.yml of mdm-core, the ApiPropertyService was updated to not allow a maximum file size configuration to be added that is greater than the Grails config maximum request size.

Bootstrap was also updated to set the api property, if it doesn't already exist, to a value that equals the Grails configured maximum file size.

### DIFF
--- a/mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/admin/ApiPropertyController.groovy
+++ b/mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/admin/ApiPropertyController.groovy
@@ -17,6 +17,7 @@
  */
 package uk.ac.ox.softeng.maurodatamapper.core.admin
 
+import org.springframework.beans.factory.annotation.Value
 import uk.ac.ox.softeng.maurodatamapper.api.exception.ApiInternalException
 import uk.ac.ox.softeng.maurodatamapper.core.controller.EditLoggingController
 import uk.ac.ox.softeng.maurodatamapper.security.User
@@ -37,6 +38,10 @@ class ApiPropertyController extends EditLoggingController<ApiProperty> {
     ApiPropertyController() {
         super(ApiProperty)
     }
+
+    @Value('${grails.controllers.upload.maxFileSize}')
+    private Integer maxFileUploadSize;
+
 
     @Override
     def index(Integer max) {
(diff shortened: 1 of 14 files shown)


mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/admin/ApiPropertyController.groovy
```diff
@@ -38,6 +39,10 @@ class ApiPropertyController extends EditLoggingController<ApiProperty> {
         super(ApiProperty)
     }
 
+    @Value('${grails.controllers.upload.maxFileSize}')
+    private Integer maxFileUploadSize;
```
joe-crawford, line 43: Hi @GBishop-PHSA,

Thanks for this contribution and sorry for the delay. There are various failing tests linked to
```
Error creating bean with name 'apiPropertyService': Unsatisfied dependency expressed through field 'maxFileUploadSize'; nested exception is org.springframework.beans.TypeMismatchException: Failed to convert value of type 'java.lang.String' to required type 'java.lang.Integer'; nested exception is java.lang.NumberFormatException: For input string: "${grails.controllers.upload.maxFileSize}"
```

It looks like there are issues using the Spring `@Value` annotation in Grails. Probably the easiest solution is to switch to `grailsApplication.config.getProperty`. Could you update so that the tests pass? Thanks.

joe-crawford (changes requested):

GBishop-PHSA:
@joe-crawford I'm not sure why the build test isn't passing: 
DataModelFunctionalSpec > MI07 : test merge into of two DMs with a Model Data Type FAILED
   java.lang.AssertionError at DataModelFunctionalSpec.groovy:2379

It passes when I run this test locally.

joe-crawford:
> @joe-crawford I'm not sure why the build test isn't passing: DataModelFunctionalSpec > MI07 : test merge into of two DMs with a Model Data Type FAILED java.lang.AssertionError at DataModelFunctionalSpec.groovy:2379
> 
> It passes when I run this test locally.

Thanks for updating, that looks like a StaleStateException error so isn't related.

I've rebased onto develop, the new failing tests are also not related. This looks good to me, I've commented on the UI PR.

mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/admin/ApiPropertyController.groovy
GBishop-PHSA: @joe-crawford I've finished updating the tests to pass.

joe-crawford:
Hi @GBishop-PHSA, many thanks for the PR, have tested successfully locally with your UI branch and seen the message when an overly large attachment is attempted.

joe-crawford (approved):

PR #433, opened 2023-08-30
Fix adding model security on async import

### LINKED ISSUE #432: Error adding security to model on async import
# Description

When an ImporterProviderService imports a single model, an error adding security to the model occurs when using the Importer via an AsyncJob.

# Environment

- mdm-core latest develop
- mdm-plugin-database-oracle import

joe-crawford:
Closes #432

Also adds indexes to speed up validating large EnumerationTypes


jamesrwelch (approved):
Good spot!  I guess ideally the indexes would go in a separate PR, but I'm happy and we're busy, so let's go :)

PR #448, opened 2024-02-02
Add model version tag to some of the views they are missing from

jamesrwelch:
Part of a collection of updates to provide model version tags in the user interface


joe-crawford (approved):
Great, have tested with the UI, looks good to me :)

PR #453, opened 2024-04-29
Add database indexes to speed up finalisation of large items

jamesrwelch:
This PR adds new database indices which, amongst other things, speeds up the finalisation of large models and versioned folders


joe-crawford:
Looks good to me 👍

Issue #355, opened 2022-08-13
'Extends' Data Classes are not copied when creating a new branch


jamesrwelch:
A Data Class that 'extends' another data class within the same model is not being copied correctly when branching / creating a new version - the 'extends' property is not being copied.

jamesrwelch:
Probably means #242 is not complete / correct

PR #454, opened 2024-04-29
Optimise the deletion of versioned folders by reducing the number of queries

jamesrwelch:
This PR massively speeds up the deletion of Versioned Folders by reducing the number of queries and not building the breadcrumb tree white so often

### DIFF
--- a/mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/container/FolderService.groovy
+++ b/mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/container/FolderService.groovy
@@ -160,7 +160,7 @@ class FolderService extends ContainerService<Folder> {
     Set<SecurableResource> delete(Folder folder, boolean permanent, boolean flush = true, boolean rebuildSecurity = true) {
         if (!folder) {
             log.warn('Attempted to delete Folder which doesnt exist')
-            return
+            return []
         }
         if (permanent) {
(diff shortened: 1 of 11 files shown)


jamesrwelch:
This also includes a change to fix #355 - sorry, was going to include that in a separate PR but it's quite related.

joe-crawford:
Hi @jamesrwelch, there's a number of failing tests on this branch that look like they may be related? E.g. tests under `ModelUserAccessPermissionChangingAndVersioningFunctionalSpec` which also are failing locally for me, I think these may need double checking if they're OK or not?

jamesrwelch:
All tests are now passing.  But something odd has happened to my local instance and I want to check it's not related - so please don't merge this one quite yet!

jamesrwelch:
Ok, it's my mistake - I hadn't rebased on top of develop.  This one's good to go.

jamesrwelch:
(Note the force-push)

pjmonks (approved):
I think this change looks good. From my understanding you're avoiding the final flush and security rebuild until the end when the top level folder is finally removed.

I've tested this before on my local instance and can definitely see the speed improvement, plus it works as I expect. I approve this.

jamesrwelch:
> From my understanding you're avoiding the final flush and security rebuild until the end when the top level folder is finally removed.

Yes - doing it after removing each model and folder in the Versioned Folder was unnecessarily slow

PR #465, opened 2024-05-23
gh-457: New endpoints for copying models

### LINKED ISSUE #457: Add endpoints for copying Mauro models and items
# Description

Mauro has some abilities to copy catalogue items, yet it does not provide enough public HTTP endpoints to carry these operations out. These need to be included to support the copying of items - this will make it easier to create new items based on more complex items which already contain a lot of information.

The following domain types must be supported:

- Data Models
- Terminology
- Term
- Code Set

**Note:** Data Classes and Data Elements are also required to fulfil a use case, but endpoints already exist for these types.

# Scope

The copy actions should only be allowed for models or sub-items which are within a draft versioned folder. It would not make sense for a model to be allowed to be copied in a finalised versioned folder - this can just be forked. This copy action represents a user wanting to copy an existing catalogue item to base on and then modify.

# URL Mappings and Services

Each item service e.g. `DataModelService` already has a `copyModel()` method (or similar), they just need to be mapped to controllers and endpoints.

The interceptors of each controller should be modified so that if a copy operation is requested, check that the current user is allowed to edit the target model/item and read the source model/item.

# Available Actions

The `UserSecurityPolicyManager` contains methods to fetch the available actions allowed for an item. The `copy` action should be included in this list as appropriate. This will assist the frontend for deciding which options are shown to the user.

`copy` should only be allowed when a model (or sub-item) is within a draft versioned folder.

# Asynchronous action

Compared to a new fork, Mauro should allow a copy to be asynchronous if requested as a parameter.

mhocc:
Resolves #457 

Initial draft of a new endpoint for copying a data model. This is WIP, more endpoints will eventually appear, but focusing on data models for now.

# Background context

Assume that users for NHS data dictionary will want to copy an existing catalogue item (e.g. a terminology of national codes, a data set) so that they can base a new item on the original and modify, rather than create entirely from scratch. This means adding additional endpoints to support the domains.

Some other restrictions that have been agreed between @pjmonks and @jamesrwelch:

1. A model can only be copied when it is contained within a versioned folder - otherwise it should just be forked
2. A source model (and extension the parent versioned folder) cannot be finalised, it must be draft. Again, it would be better to fork otherwise.
3. A new copied model must be targeted to a sub-folder within the same parent versioned folder as the source model - it cannot be copied outside of this boundary

# Completed

- `ModelController` updated with `copyModel()` action
- URL mappings for data models currently in place to use `ModelController.copModel()`
- Checks in place to ensure that inputs are valid and preconditions are met (see above)
- Functional tests added to test everything so far
- Copying terms 
- Copying terminologies
- Copying CodeSets
- Asynchronous copying for sub-classes of DataModels

Not sure how to fully test async copying, needs some method of finding the copied item—by label?

The following need checking to see if it is applying the correct restrictions in the controller interceptors:
-   The TODO in `ModelInterceptor.checkModelActionsAuthorised()`
-   The TODO in `TermInterceptor.before()`

The copying of `DataClass` and `DataElement` types appear to have been implemented—no additional work done for these.


jamesrwelch:
@pjmonks I think this looks like a sensible approach, yes 👍

mdm-plugin-terminology/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/terminology/item/TermInterceptor.groovy
```diff
@@ -49,6 +49,11 @@ class TermInterceptor extends TerminologySecuredInterceptor {
             return canReadModel()
         }
 
+        // TODO: is this appropriate?
+        if (actionName == 'copyTerm') {
+            return canEditModelAndReadOtherModel()
```
pjmonks: This isn't quite right, but the intention is correct. The URL mapping and expected request body (JSON) is this:

```
PUT /terminologies/{terminologyId}/terms/copy/{termId}

{
    "targetTerminologyId": "uuid",
    "label": "some label"
}
```

Any variables in the URL route are found in the `params` map, and anything in the request body can be found in the `request` map (I think...)

When `actionName` is `"copyTerm"`, it will need to check:

1. Can the source terminology be read? I think this is done by `currentUserSecurityPolicyManager.userCanReadSecuredResourceId(Terminology.class, params.terminologyId)`
2. Can the target terminology be edited? I think this is done by `currentUserSecurityPolicyManager.userCanEditSecuredResourceId(Terminology.class, request.targetTerminologyId)`

If either of those are `false` then return `notFound()`. It would be helpful if @joe-crawford could confirm I got this correct too.

mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/interceptor/ModelInterceptor.groovy
```diff
@@ -117,6 +117,15 @@ abstract class ModelInterceptor extends TieredAccessSecurableResourceInterceptor
             return checkExportModelAction()
         }
 
+        // TODO: check this before merge - pure guesswork
+        if (actionName == 'copyModel') {
+            boolean canReadModel = currentUserSecurityPolicyManager.userCanReadSecuredResourceId(getSecuredClass(), getId())
```
pjmonks, line 121: This seems correct to me 👍

pjmonks (changes requested):
This is looking great to me! I've tested manually that I can copy a data model and terminology and this works. I wasn't able to test copying a term yet since the interceptor security is not letting me - please review my comment on that and fix.

The automated tests all look good too.

One question for @jamesrwelch and @joe-crawford - do these changes require a new available action to be returned for each resource in the `availableActions` array, such as `copy`? From previous experience fiddling with this that sounds complex and breaks lots of tests. Or could we make do with one of the existing ones, such as `save`?

joe-crawford:
> This is looking great to me! I've tested manually that I can copy a data model and terminology and this works. I wasn't able to test copying a term yet since the interceptor security is not letting me - please review my comment on that and fix.
> 
> The automated tests all look good too.
> 
> One question for @jamesrwelch and @joe-crawford - do these changes require a new available action to be returned for each resource in the `availableActions` array, such as `copy`? From previous experience fiddling with this that sounds complex and breaks lots of tests. Or could we make do with one of the existing ones, such as `save`?

Hi @pjmonks , yes aware how painful that is, definitely if we can piggy back on an existing permission/action, makes sense. Can anyone who can read the model create a copy? In that case doesn't even really need to be listed specifically?

mdm-plugin-terminology/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/terminology/item/TermInterceptor.groovy
joe-crawford: @pjmonks Yes definitely agree with your logic above, and also agree that the canEditModelAndReadOtherModel() call won't work, the model IDs will be the wrong way round and `targetTerminologyId` won't be picked up.

mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/interceptor/ModelInterceptor.groovy
joe-crawford: In ModelInterceptor, there's also this bit 

```
            // We control addition of models into containers by using container permissions
            if (isSave()) {
                return currentUserSecurityPolicyManager.userCanCreateResourceId(Model, null, Folder, params.folderId) ?:
                       forbiddenOrNotFound(canReadFolder, Folder, params.folderId)
            }
```

which checks the parent folder is writeable, should also do a similar check when copying to a new folder. Probably can tweak this to be `isSave() || isCopy()`.

pjmonks:
> > This is looking great to me! I've tested manually that I can copy a data model and terminology and this works. I wasn't able to test copying a term yet since the interceptor security is not letting me - please review my comment on that and fix.
> > The automated tests all look good too.
> > One question for @jamesrwelch and @joe-crawford - do these changes require a new available action to be returned for each resource in the `availableActions` array, such as `copy`? From previous experience fiddling with this that sounds complex and breaks lots of tests. Or could we make do with one of the existing ones, such as `save`?
> 
> Hi @pjmonks , yes aware how painful that is, definitely if we can piggy back on an existing permission/action, makes sense. Can anyone who can read the model create a copy? In that case doesn't even really need to be listed specifically?

The key part of copying is that a user is able to create a new model/item in the target location, so it's effectively the same permission as "create" (or whatever that is called). So let's ignore an additional action then.

mdm-plugin-terminology/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/terminology/item/TermInterceptor.groovy
mhocc: Implemented Pete's suggestion

mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/interceptor/ModelInterceptor.groovy
mhocc: Not sure if the suggested change belongs to this task.

pjmonks (approved):
Had to make a tweak to the `TermInterceptor` security again but my manual testing shows it works now. Ready to go ✔️

mdm-core/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/core/interceptor/ModelInterceptor.groovy
joe-crawford: Noted that it looks like there's a restriction to only copy models within their original versioned folder, however is there any check that the user can actually edit the versioned folder rather than just read it?

mdm-plugin-terminology/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/terminology/item/TermInterceptor.groovy
```diff
@@ -49,6 +52,26 @@ class TermInterceptor extends TerminologySecuredInterceptor {
...
+            //  the controller. Figure out how to solve this later
+//            DataBindingSource bindingSource = cacheRequestBody()
+//            UUID targetTerminologyId = bindingSource && bindingSource.containsProperty('targetTerminologyId')
+//                ? UUID.fromString(bindingSource.getPropertyValue('targetTerminologyId').toString())
+//                : null
+//
+//            boolean canReadTarget = currentUserSecurityPolicyManager.userCanReadSecuredResourceId(Terminology, targetTerminologyId)
+//            if (!currentUserSecurityPolicyManager.userCanEditSecuredResourceId(Terminology, targetTerminologyId)) {
+//                return forbiddenOrNotFound(canReadTarget, Terminology, targetTerminologyId)
+//            }
```
joe-crawford: If it isn't possible to get the `targetTerminologyId` out in the Interceptor, it should be possible to just add the call to `currentUserSecurityPolicyManager.userCanEditSecuredResourceId(Terminology, targetTerminologyId)` directly within the controller.

pjmonks:
@joe-crawford I've added some additional security checks you requested, please can you check they're OK?

pjmonks:
Only one test failure this time, again a different test and I think due to a database error when running, so I don't think it's related to this change.

joe-crawford:
Great, thanks @pjmonks, @mhocc, looks good to me :+1:

PR #469, opened 2024-06-27
Move Data Element between Data Classes

### LINKED ISSUE #468: Allow updates to Data Elements to change the parent Data Class
# Description

There is a requirement to be able to move a Data Element to a different Data Class within the same Data Model. At the moment, the `PUT /api/dataModels/{modelId}/dataClasses/{classId}/dataElements/{elementId}` will fail if a different `dataClass` parent UUID is included in the request body.

# Steps to reproduce

Send this request:

```http
PUT /api/dataModels/{modelId}/dataClasses/{classId}/dataElements/{elementId}

{
    "dataClass": "{anotherClassId}"
}
```

Note that the `model` property is not included in the body - this should still not be allowed, since moving a data element outside the model to a different model would cause links to the data type to break.

This is the exception currently thrown:

```
deleted object would be re-saved by cascade (remove deleted object from associations): [uk.ac.ox.softeng.maurodatamapper.core.facet.BreadcrumbTree#c665b724-fed0-4918-8197-d78286034371]
```

# Expected behavior

A `200 OK` should return in the response. Once complete, the data element should now appear in Mauro under it's new data class parent.

# Related

- NHSDigital/DataDictionaryPublication#468

jamesrwelch:
Code to move a data element from one class to another.  E.g.
`PUT api/dataModels/{dataModelId}/dataClasses/{dataClassId}/dataElements/{dataElementId}`
with body:
```json
{
    "dataClass": {
        "id": "b2ce91df-2606-46d0-a98b-b310e5d1f1be"
    }
}
```
Other properties (label, description) can be changed at the same time.

There's one failing test which I believe to be unrelated

### DIFF
--- a/mdm-plugin-datamodel/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/datamodel/item/DataElementController.groovy
+++ b/mdm-plugin-datamodel/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/datamodel/item/DataElementController.groovy
@@ -27,10 +27,12 @@ import uk.ac.ox.softeng.maurodatamapper.datamodel.item.datatype.DataType
 import uk.ac.ox.softeng.maurodatamapper.datamodel.item.datatype.DataTypeService
 
 import grails.gorm.transactions.Transactional
+import groovy.util.logging.Slf4j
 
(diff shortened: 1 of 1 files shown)


joe-crawford:
Looks good to me! 👍

PR #471, opened 2024-06-28
Add filtering to endpoint that provides DataElements per DataModel

### LINKED ISSUE #470: Allow filtering on the "List all data elements in a data model" endpoint
# Description

The [List all data elements in a data model](https://documenter.getpostman.com/view/9840589/UVC8BkkA#0342ed1f-262f-4b65-8274-77cb13133fbd) endpoint currently only supports pagination and sorting but does not support filtering. Given it is possible that the returned list will contain a total of hundreds of items, it seems important that filters should be supported.

The following at least should be allowed as filters:

- Filter by data element label

# Additional context

This relates to:

- MauroDataMapper/mdm-ui#855

The UI could only be partially completed due to missing filter options.

jamesrwelch:
One failing test which I think is unrelated

Resolves #470

### DIFF
--- a/mdm-plugin-datamodel/grails-app/services/uk/ac/ox/softeng/maurodatamapper/datamodel/item/DataElementService.groovy
+++ b/mdm-plugin-datamodel/grails-app/services/uk/ac/ox/softeng/maurodatamapper/datamodel/item/DataElementService.groovy
@@ -310,8 +310,26 @@ WHERE (de.dataClass.id = :dataClassId OR idc.id = :dataClassId)''', 'de', filter
         DataElement.byImportingDataClassIdInList(dataClassIds).list()
     }
 
-    List<DataElement> findAllByDataModelId(Serializable dataModelId, Map pagination = [:]) {
-        DataElement.byDataModelId(dataModelId).list(pagination)
+    List<DataElement> findAllByDataModelId(Serializable dataModelId, Map filters = [:], Map pagination = [:]) {
+        Map<String, Object> queryParams = [dataModelId: dataModelId]
+
+        String baseQuery = applyHQLFilters('''
+FROM DataElement de
+WHERE (de.dataClass.dataModel.id = :dataModelId)''', 'de', filters)
+
+        // Cannot sort DEs including imported using idx
+        String sortedQuery = applyHQLSort(baseQuery, 'de', pagination[ARGUMENT_SORT] ?: ['label': 'asc'], pagination, true)
+
+        new HQLPagedResultList<DataElement>(DataElement)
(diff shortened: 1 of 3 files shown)


mdm-plugin-datamodel/grails-app/services/uk/ac/ox/softeng/maurodatamapper/datamodel/item/DataElementService.groovy
```diff
@@ -310,8 +310,26 @@ WHERE (de.dataClass.id = :dataClassId OR idc.id = :dataClassId)''', 'de', filter
         DataElement.byImportingDataClassIdInList(dataClassIds).list()
     }
 
-    List<DataElement> findAllByDataModelId(Serializable dataModelId, Map pagination = [:]) {
-        DataElement.byDataModelId(dataModelId).list(pagination)
+    List<DataElement> findAllByDataModelId(Serializable dataModelId, Map filters = [:], Map pagination = [:]) {
+        Map<String, Object> queryParams = [dataModelId: dataModelId]
```
pjmonks, line 315: I've spotted a problem when testing this locally. I tried `GET /dataModels/{dataModelId}/dataElements?max=20&label=accomodation` to test filtering the data element label but got a 500 response saying "Named parameter not bound : label" from Hibernate.

I compared to the data type list endpoint `GET /dataModels/{dataModelId}/dataTypes?max=20&label=abdom` which works. Comparing the `DataTypeService`, it has this line in `findAllDataTypesByHQLQuery()`:

```groovy
queryParams.putAll(extractFiltersAsHQLParameters(filters))
```

This isn't included in the `DataElementService`. So I think it's still only filtering on `dataModelId` but the HQL query now includes a `:label` parameter it cannot resolve.

pjmonks (changes requested):
Seems close, but returns an error for me, please review my comment.

jamesrwelch:
@OButlerOcc @pjmonks  I believe this is now ready for re-review.  Note the force-push after rebasing on develop.

pjmonks (approved):
I've tested your fix again and it works now. I've also pushed up my own addition, I figured out how to filter by both label for the data element and the parent data class too. If any tests fail on Jenkins I believe they will be unrelated. Ready to merge ✔️

joe-crawford:
@pjmonks, @jamesrwelch FYI I've tried running the failing tests locally, `FolderTreeItemFunctionalSpec` passed but `VersionedFolderFunctionalSpec` failed at the same test (DI06) as in Jenkins.

joe-crawford:
VersionedFolderFunctionalSpec DI06 is passing locally on develop so think something has been affected there. Not sure exactly what though.

PR #476, opened 2024-07-15
Check for administrators group on Bootstrap

jamesrwelch:
Inelegant but effective solution for NHS DD issue after renaming administrator group


pjmonks (approved):
These changes look fine to me. I've noticed that Jenkins has failed, the 3 failed tests don't look related to this change (otherwise I would have expected more to fail) but maybe worth double checking locally. I suspect they are fine to ignore though.

jamesrwelch:
@pjmonks Agreed the failing test is unrelated - I re-ran and a different test failed.  I think this is ready to go

Issue #479, opened 2024-07-16
Merging versioned folders that are contained in sub-folders cannot find child items by path


pjmonks:
# Description

I've discovered that to successfully merge a versioned folder branch into another, both must be at the top-level of the catalogue tree. If one or both are under a sub-folder, then locating the source/target items relative to the source/target versioned folders does not work because the `PathService` cannot find the items.

# Steps to reproduce

First, create two versioned folders with branches `main` and `anotherBranch`. Make sure that `main` already contained some models/items, then add some new models/items to `anotherBranch`. To verify the merge works generally, first create them at the root level of the catalogue:

![image](https://github.com/user-attachments/assets/fa5f3b24-0d90-43b4-900d-10d278f7d9d3)

Perform a merge from `anotherBranch` into `main`. Once completed, this should successfully work and the two branches should be identical.

Now move these versioned folders to a sub-folder in the catalogue tree:

![image](https://github.com/user-attachments/assets/c38bc491-2667-4af2-aa89-2cfff401b7fd)

Now merge `anotherBranch` into `main` again, but this time the UI says it completed successfully but the new items were not copied from source to target.

# Expected behavior

When under sub-folders, the versioned folders should still be identical after a merge.

# Additional context

In terms of creation patches, I think the problem is because the path is absolute to the catalogue, not the source/target versioned folder. This is an example of a new item created in a source branch under a sub-folder:

```
fo:Test Folder|vf:Test Versioned Folder$anotherBranch|te:Versioned Terminology$anotherBranch
```

The `PathService.findResourceByPathFromRootResource()` method will look at this, see it is not relative to the provided source versioned folder, and attempt to use the whole path for item lookup. This will result in using the `FolderService` to locate the item, but this the wrong level to look at.

I think this path should be truncated during the merge - all sub-folders before `vf:Test Versioned Folder$anotherBranch` should be disregarded so that the child path can be used correctly.

PR #480, opened 2024-07-16
gh-475 Edit history for merged items under versioned folders

### LINKED ISSUE #475: Record the merging of a versioned folder into the individual target items
# Description

When a versioned folder is merged into a target versioned folder, currently an overall commit message is recorded on the target VF to represent the change history.

This should be extended to add a history record to each individual item that was changed in the target VF too. This will highlight at a more granular level how the merge affected the VF.

# Merge operation

The `VersionedFolderService.mergeObjectPatchDataIntoVersionedFolder()` method has three top level operations to take care of for every patched item/field change:

- Created items in the target
- Deleted items from the target
- Modified items in the target

For each one, a new history record should be added as each catalogue item is changed/created using the `EditHistoryAware` trait. The `EditTitle.MERGE` enum should be used along with a suitable description referencing the source versioned folder that caused the change, e.g.

- "Item created in branch 'my-test-branch'"
- "Item modified in branch 'my-test-branch'"
- etc

# Created and updated items

As items are created (copied) or modified in the target versioned folder, the history should be added directly to that new/updated item in question.

# Deleted items

Where items have been deleted from the target versioned folder, the history item should be added to the parent item that used to own the child deleted item.

pjmonks:
Resolves #475 

Track at an item level changes that have happened to items under a versioned folder during a merge:

- Created/modified items have a `MERGE` edit history record attached to them.
- Deleted items have a `MERGE` edit history record attached to their parent (or the target versioned folder if not available).

This then results in a visible change history at the item level, as well as the `CHANGELOG` entry recorded on the versioned folder itself.

image

I have adjusted the `mdm-testing-functional` tests for Versioned Folder merging to carry out some spot checks to ensure that these `MERGE` history entries are at least being recorded.

### DIFF
--- a/mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/container/VersionedFolderService.groovy
+++ b/mdm-core/grails-app/services/uk/ac/ox/softeng/maurodatamapper/core/container/VersionedFolderService.groovy
@@ -56,6 +56,7 @@ import uk.ac.ox.softeng.maurodatamapper.core.rest.converter.json.OffsetDateTimeC
 import uk.ac.ox.softeng.maurodatamapper.core.rest.transport.merge.FieldPatchData
 import uk.ac.ox.softeng.maurodatamapper.core.rest.transport.merge.ObjectPatchData
 import uk.ac.ox.softeng.maurodatamapper.core.rest.transport.model.VersionTreeModel
+import uk.ac.ox.softeng.maurodatamapper.core.traits.domain.EditHistoryAware
 import uk.ac.ox.softeng.maurodatamapper.core.traits.domain.MultiFacetItemAware
 import uk.ac.ox.softeng.maurodatamapper.core.traits.service.MdmDomainService
 import uk.ac.ox.softeng.maurodatamapper.core.traits.service.MultiFacetItemAwareService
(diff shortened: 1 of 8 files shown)


pjmonks:
I have pushed some fixes due to failed tests, part of my changes broke parameter lists and overridden methods in the `ModelService`. All the merge tests pass now, but one new test has failed for `ProfileFunctionalSpec` in `mdm-testing-functional`. This originally passed before and passes locally on my machine, I'm hoping this is an outlier.

jamesrwelch:
The code changes look good and the tests look sensible.  I've just tried it myself, on a minimal versioned folder, and I'm not getting any `MERGE` edits for the PrimitiveType that I've modified.  Have I done something wrong?

pjmonks:
@jamesrwelch I've found the cause of your issue, the data type is recording the edit history during the merge correctly, as you can see in Postgres:

image

The problem is actually the UI, it's requesting a resource type of `dataTypes` but needs to be more specific like `primitiveTypes`, so it is not fetching the history correctly.

Here are the current known issues related to merging and the change history:

- #479 
- MauroDataMapper/mdm-ui#889
- MauroDataMapper/mdm-ui#890

jamesrwelch (approved):
I've had a play and it seems to work nicely - the code changes look good too.  Well done for making this fairly major change so quickly!  
I agree the failing test in Jenkins is unrelated
With the caveat that the other issues reported will also need fixing, I'm very happy with this 👍

PR #483, opened 2024-07-23
gh-482 fixed history totalcount bug

### LINKED ISSUE #482: Support for history totalcount bug mdm-ui/feature/gh-890
support for: https://github.com/MauroDataMapper/mdm-ui/compare/feature/gh-890---pagination-in-history-tab?expand=1

Calls to the edit controller do not return the correct totalCount value regardless of what pagination parameters used

OButlerOcc:
Supports gh-890 on mdm ui

closes #482 

The history tab was not receiving the correct totalcount of records when requesting the history of an item. This caused the paginator to not function at all. 
This was caused by two problems. 
First the call to access the database used the gorm getAllByX notation, which does not return the paginatedResponse object.
Second, the controller was forming a collect function on the list to populate the users, which was bricking the totalCount by converting the PagedResponseObject into a List object.
This caused the gson view to see default to the non paged response format and return the call as a simple list with count = to the current query count.

Solution was to refactor the call to a detachedCritera call which does return a pagedResponseObject (if you know why this is a bad idea please tell me, its been used elsewhere but I'm a little wary of the word detached)

And to modify the objects by value rather than by reference and return the original pagedResponseObject, preserving the count.

Testing:
see 
gh-890 on mdm ui

### DIFF
--- a/mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/facet/EditController.groovy
+++ b/mdm-core/grails-app/controllers/uk/ac/ox/softeng/maurodatamapper/core/facet/EditController.groovy
@@ -20,6 +20,7 @@ package uk.ac.ox.softeng.maurodatamapper.core.facet
 
 import uk.ac.ox.softeng.maurodatamapper.core.traits.controller.MdmController
 
+import grails.gorm.PagedResultList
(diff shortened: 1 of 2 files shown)


OButlerOcc:
wrong branch name, closed and remade

PR #486, opened 2024-07-25
gh-479 Merge Versioned Folders that are under sub-folder in the catalogue

### LINKED ISSUE #479: Merging versioned folders that are contained in sub-folders cannot find child items by path

pjmonks:
Resolves #479 

# Overview

If you structured a Versioned Folder at the root of the catalogue, it would be able to be merged into another root level Versioned Folder. But, if the VFs were under sub-folders themselves, the paths would be absolute to the catalogue root and then resources like models would not be located through the `PathService` at all.

I've fixed this by reducing all paths used in the merge down to the source/target locations of the versioned folders, removing all ancestor path nodes until the expected VF paths are reached. For example, if a creation patch had this path:

```
fo:Testing Folder|fo:Branches|vf:Test Versioned Folder$another-branch|te:Test Versioned Terminology$another-branch
```

Where the Versioned Folder source was `fo:Testing Folder|fo:Branches|vf:Test Versioned Folder$another-branch`, then the creation source path is reduced to:

```
vf:Test Versioned Folder$another-branch|te:Test Versioned Terminology$another-branch
```

This means the `PathService.findResourceByPathFromRootResource()` method will work with this path.

- Updated `Path` to include a `reduce()` method
- Updated the unit tests of `Path` to verify `reduce()` works

# Testing

There are no new automated tests, but the existing tests still pass.

I tested manually by creating some VFs in this structure - note that I created multiple sub-folders to ensure that the paths could be reduced correctly:

image

Create/modify/delete from the other branch and merge into the `main` branch - this should then correctly work.

### DIFF
--- a/mdm-common/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/path/Path.groovy
+++ b/mdm-common/src/main/groovy/uk/ac/ox/softeng/maurodatamapper/path/Path.groovy
@@ -148,10 +148,10 @@ class Path implements Serializable, Cloneable {
         getPathString(modelIdentifierOverride)
     }
 
-    Path clone() {
+    Path clone(String modelIdentifierOverride = null) {
         Path local = this
         new Path().tap {
-            pathNodes = local.pathNodes.collect {it.clone()}
+            pathNodes = local.pathNodes.collect {it.clone(modelIdentifierOverride) }
         }
     }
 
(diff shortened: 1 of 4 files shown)


OButlerOcc (approved):
Tested fine with a variety of subfolders and changes, breakpointing through the code shows that the reduce function is working fine. An optional change would be to update the docs around the reduce function to make it a bit clear why its doing what its doing

pjmonks:
@joe-crawford or @jamesrwelch - another approved pull request with failing tests. The two Jenkins builds give different build errors, 1 test is failing on each one (but different ones), and they pass locally. Please could this be merged?

